PR #34, opened 2015-03-12
add optional parameter to init() in order to suppress output

fietew:


fietew:
Since mxIsLogical does not accept 0/0.0 or 1/1.0 as input parameters I modified the conversion function

PR #37, opened 2015-06-23
change source port xml request revised

c-bebop:
In order to change the jackport I figured to use:
```<request><source id="42" port="system:capture_3"/></request>```
Instead of:
```<request><source id="42" port_name="system:capture_3"/></request>```


mgeier:
Thanks!

PR #43, opened 2016-06-06
help2man dependency for ubuntu

mgeier:
help2man is missing, compiling aborts. Installing fixes it. Tested on Ubuntu 14.04 and 15.10.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -127,7 +127,8 @@ dnl AC_DEFINE writes values in config.h
 dnl AC_DEFINE_UNQUOTED performs additional shell expansions, e.g. substitution
 dnl of variable with value.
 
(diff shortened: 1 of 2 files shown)


mgeier:
I've added 4e46ce4bd9c492aaf44758bf0705eda3c406a8ba, does this work for you?
Some kind of warning message would probably be nice (if `help2man` is missing), but I don't know how to do that. Any ideas?

chohner:
Just to chime in: The same is true for building on OSX el capitan. Building the manual fails even with `help2man` installed. I almost got it working by removing the '--locale=en' flag in 'man/Makefile.am' - I will have to check what made me give up at the very end.

umlaeute:
@chohner so does [this pull request]() or [mine](#44) fix the issue for you?

PR #44, opened 2016-06-06
print warning if help2man is not found

umlaeute:
warning, if help2man is missing - SoundScapeRenderer/ssr#43


mgeier:
Thanks a lot!

I've merged this into `master`, but Github somehow didn't seem to grok that and marked this as "closed" instead of "merged" ...

PR #50, opened 2016-08-24
Fix OSX build process

chohner:
- adds -f to both cp (ecasound and qt_menu.nib) to not fail if the file already exists
- sets the necessary -w permission on the copied qt frameworks so install_name_tool can do it's thing


chohner:
Some motivations for this PR:

https://github.com/SoundScapeRenderer/ssr/pull/50/commits/6536b3b491362fab7df31c7eac37e2abd63d6f73 helps re-building the dmg image if a previous build has failed at a later stage, since cp without the `-f` flag errors out if it finds `ecasound` and/or `qt_menu.nib` has already already copied over before.

https://github.com/SoundScapeRenderer/ssr/pull/50/commits/22b578dc059685128c0a3d943b304d79a16162f2 helps makes all copied over frameworks writeable, so `install_name_tool` can change them without `sudo` (currently one needs to run `sudo make dmg`).

Issue #52, opened 2016-11-02
git describe in configure.ac fails when building from an archive/release


FRidh:
[This line](https://github.com/SoundScapeRenderer/ssr/blob/0dd0136dd24e47b63d8a4e05de467f5c7b047ec9/configure.ac#L19) is problematic.  When git is not available it will fail. Furthermore, when installing from an archive (e.g. a release) and git is available it will fail because the directory is not a repository. I now circumvent this by creating an empty repo with `git init` which results in the git warning `fatal: No names found, cannot describe anything.`.

umlaeute:
I faced that very problem for the Debian package.
The solution i ended up with, was to replace the `git describe` line with a shell-script that gets the version from "somewhere else" (in my case: from the Debian packaging metainformation).

https://anonscm.debian.org/cgit/pkg-multimedia/soundscaperenderer.git/tree/debian/patches/upstreamversion.patch

mgeier:
Thanks for the report.

But does it really fail?

AFAICT, it shows a disturbing message but it should still work. If `git` is not available or the call to `git` fails for some reason, a fallback "version" string will be created from the current user name and a date/time string. That's what the part after the `||` is supposed to do:

    git describe || date +$USER%Y%m%d%H%M

Anyway, I've created #53 which ignores the error output by using this:

    git describe 2> /dev/null || date +$USER%Y%m%d%H%M

Can you please check if that works for you?

FRidh:
Yes, that fixes it. Thanks.

PR #53, opened 2016-11-02
Ignore error message if git is not available (or fails)

mgeier:


FRidh:
As I wrote in https://github.com/SoundScapeRenderer/ssr/issues/52 this fixes the issue.

Issue #54, opened 2016-11-03
help2man: can't get `--help' info from ../src/ssr-binaural


FRidh:
While building on Nix, I encountered the following issue

```
Making all in man
make[1]: Entering directory '/tmp/ssr/ssr-0dd0136/man'
help2man --no-info --name="manual page for the `test binaural = binaural -o binaural = generic && echo binaural || echo binaural | tr 'a-z' 'A-Z'` renderer" --output=ssr-binaural.1 --locale=en_US \
        ../src/ssr-binaural
help2man: can't get `--help' info from ../src/ssr-binaural
Try `--no-discard-stderr' if option outputs to stderr
make[1]: *** [Makefile:568: ssr-binaural.1] Error 127
make[1]: Leaving directory '/tmp/ssr/ssr-0dd0136/man'
make: *** [Makefile:549: all-recursive] Error 1
```

Note that I've patched the locale to `en_US` since `en` isn't available/valid.
cc @umlaeute @mgeier

FRidh:
Solved. `help2man` basically runs `./ssr-binaural  --help` and `ssr-binaural` needs `libGL`.

mgeier:
OK, cool, do you still need the change to `en_US`?

FRidh:
Yes, that change is needed.

mgeier:
Can you please make a PR?

PR #55, opened 2016-11-04
Change default locale help2man

FRidh:
`en` is not a valid locale, at least, not on my system (`locale -a`).
Let's use `en_US` instead.

See also https://github.com/SoundScapeRenderer/ssr/issues/54.

### DIFF
--- a/man/Makefile.am
+++ b/man/Makefile.am
@@ -21,10 +21,14 @@ RENDERER_TEXT = "manual page for the \
 # Note the order of SUBDIRS in the top-level Makefile.
 #
 # "make" must be called before "make dist", there is no automatic dependency!
+HELP2MAN_LOCALE ?= $(shell locale -a | egrep "C.UTF-8" || echo "C")
 
-$(dist_man_MANS): $(top_srcdir)/src/configuration.cpp $(top_srcdir)/configure.ac
-	help2man --no-info --name=$(RENDERER_TEXT) --output=$@ --locale=en \
-		$(top_builddir)/src/$(@:.1=$(EXEEXT))
+$(dist_man_MANS): $(top_srcdir)/src/configuration.cpp $(top_srcdir)/configure.ac 
+	help2man \
+		--no-info \
+                --name=$(RENDERER_TEXT) \
+                --output=$@ \
+                --locale=$(HELP2MAN_LOCALE) $(top_builddir)/src/$(@:.1=$(EXEEXT))
 
 ## Settings for Vim (http://www.vim.org/), please do not remove:
 ## vim:textwidth=80:comments+=bO\:##


umlaeute:
why not use the posix-locale `C` instead?

mgeier:
@umlaeute I just tried this and the umlauts were not displayed correctly in the manpage (they were shown as `??`).

FRidh:
This was the original warning I got:
```
sh: warning: setlocale: LC_ALL: cannot change locale (en)
```
Umlauts weren't presented correctly in the man pages, but they are when viewing the result of the -V option or in any of the (UTF-8) source files. Unfortunately changing encoding to en_US.UTF-8 didn't make difference, and neither did using `C`.

umlaeute:
@mgeier `C.UTF-8` probably?

i kind of dislike using `en_US` for no apparent reason.

mgeier:
@umlaeute Yes, that works for me!

@FRidh Can you please try that, too, and if you are OK with it, change it?

FRidh:
@umlaeute I agree and think that `C.UTF-8` is a better solution. Unfortunately my distribution doesn't provide `C.UTF-8` yet (some work is needed here). 

Anyway, I would find it odd if the characters would be correctly shown with `C.UTF-8` and not with `en_US.UTF-8` since the country specific parts shouldn't, as far as I know, affect the encoding.

umlaeute:
well, *my* distribution allows to me to choose the available locales. :-)

anyhow, here's a few thought:

- the system should use *defaults* whenever possible; rather than make things explicit
- if things must be made explicit, they should be overridable via variables
- don't add hacks to fix issues caused by inconsistent desktop systems
- try to get rid of implicit assumptions based on your own system

so i think that ideally, the `--locale` flag should be omitted from the makefile.
why this doesn't work (and I *can* reproduce the issue), I don't know; it probably has to do with `man` (or `troff`) preferring ASCII.

unfortunately the `LANG` environment variable doesn't help either, so we probably *must* use the `--locale` cmdline option.

i'd  therefore suggest:

```Makefile
HELP2MAN_LOCALE=C.UTF-8
$(dist_man_MANS): $(top_srcdir)/src/configuration.cpp $(top_srcdir)/configure.ac
        help2man --no-info \
                 --name=$(RENDERER_TEXT) \
                 --output=$@ \
                 --locale=$(HELP2MAN_LOCALE) \
                 $(top_builddir)/src/$(@:.1=$(EXEEXT))
```

FRidh:
Unfortunately `C.UTF-8` isn't supported by glibc. 
https://sourceware.org/glibc/wiki/Proposals/C.UTF-8
Certain distributions added support for it themselves, so I don't think it is anything that can be relied on. 

Anyway, as @umlaeute suggested creating an env var is the easiest and most flexible solution.

mgeier:
Thanks @umlaeute for the suggestion, that's great and definitely the way to go!

Not at all expecting that it would work, I tried this variation just out of curiosity:

```make
HELP2MAN_LOCALE = .UTF-8
```

And to my own surprise it worked!

This one also worked:

```make
HELP2MAN_LOCALE = UTF-8
```

What do you think about that?

@FRidh Can you please try this on your distribution?

umlaeute:
@FRidh, re:
> the country specific parts shouldn't, as far as I know, affect the encoding

sure, but the country-specific parts have side effects.

mgeier:
@FRidh @umlaeute How should we proceed?

umlaeute:
use my suggestion :-) (and let those who don't have `C.UTF-8` override it).

or use (the known to be broken) `C`, and override it for each build.

or use this:

     HELP2MAN_LOCALE ?= $(shell locale -a | egrep "C.UTF-8" || echo "C")

(which will check whether `C.UTF-8` is available and fall back to `C`; only tested on Debian though...)

FRidh:
I agree with the solution of @umlaeute to have an environment variable `HELP2MAN_LOCALE`.
The last suggestion should always work; it won't result in nice man pages when `C.UTF-8` isn't available, but at least it is possible to override the locale with `HELP2MAN_LOCALE`.

mgeier:
@FRidh Cool, can you please change the PR accordingly?

chohner:
Just a quick feedback: On OSX 10.11, `C` is available, `C.UTF-8` and `en` are not. Using `C`, I'm still struggling to actually build the final pdf but the Umlauts in the `help2man` output look fine.

mgeier:
@chohner What PDF are you talking about? Are you trying to convert the manpages to PDF?

FRidh:
Note that I've updated the PR.

umlaeute:
btw, here's an even nicer default:

    HELP2MAN_LOCALE ?= $(shell (locale -a | egrep "^C\.UTF-8|en.*\.utf8$$" || echo "C") | sort | head -1)

it will use `C.UTF-8` and failing that fall back to *any* available English locale, and only as a last resort it will use `C`.

mgeier:
@FRidh Thanks for the update!

@umlaeute and @FRidh:

I'm getting a warning when running `./autogen.sh` (on Debian Linux):

```
man/Makefile.am:24: warning: shell locale -a | egrep "C.UTF-8" || echo "C": non-POSIX variable name
man/Makefile.am:24: (probably a GNU make extension)
```

I think all these "locale -a" heuristics are getting out of hand. Do we really need all this?

Also, on my Debian system I have locales like `en_US.utf8`, whereas on a Mac at work, they are more like `en_US.UTF-8`, which wouldn't work with the suggested grep string.

Why not just keep it simple and use (like @umlaeute suggested earlier):

```
HELP2MAN_LOCALE = C.UTF-8
```

The user would still have the possibility to overwrite this with an environment variable.

FRidh:
Fine with me as well. There is no version that works on every platform, so
we might as well just pick a value and document how to set the correct
value.

JensAhrens:
A quick reminder of what @chohner said: Neither `en` nor `C.UTF-8` are available on our Mac OS X systems. I tried a number of options including `en_US.UTF-8`. The effect was that umlauts were replaced with all sorts of characters. @FRidh 's solution in this PR makes all umlauts be printed as question marks. I find that the least unfortunate fail. I suggest we go with @FRidh 's proposal.

Remind me if that one is inconvenient on any of your systems.

dvzrv:
So, can this be merged?

mgeier:
> So, can this be merged?

There doesn't seem to be consensus yet.

My preferred option would be:

```makefile
HELP2MAN_LOCALE ?= C.UTF-8
```

JensAhrens:
Guys, what should we do with this?

FRidh:
Closing because I have no need for it anyway.

PR #57, opened 2016-11-14
C++11 threads

mgeier:
This adds a new `thread_policy` using C++11 threads and uses it for the SSR.

This should enable multithreading support for all platforms which have a C++11-compatible compiler, i.e. it should theoretically also work on Windows.


mgeier:
This should be at some point superseded by https://github.com/AudioProcessingFramework/apf/issues/7

chris-hld:

PR #58, opened 2016-11-15
Towards QT5

chohner:
This PR implements most steps necessary to switch over to QT5. QT4 support was dropped in Dec 2015 (afaik) and additionally, qt5 brings much sharper GUI on hi-res displays out-of-the-box.

This was successfully tested on OSX (included dmg building) and Ubuntu 16.10, please report your experience.
On ubuntu, I had to manually set the `-fPIC` flag while compiling, to avoid the error below:
```
"You must build your code with position independent code if Qt was built with -reduce-relocations. " "Compile your code with -fPIC (-fPIE is not enough)."
```

Configure.ac:
- pkg_check for Qt5Core, Qt5Gui, Qt5OpenGL > 5.0.0
- pkg_config Qt5Core, Qt5Gui

In src/gui/*:
- Changed <Qxxxx> includes to right submodules, eg: `<QLabel>` -> `<QtWidgets/QLabel>`
- Use `.toUtf8()` instead of deprecated `.toAscii()`
- Sorted includes by source and alphabetically

OSX installer changes:
- Copy and link Version 5 instead 4
- Add additional frameworks (QtWidgets & QtPrintSupport)
- Copy & link `libqcocoa.dylib` instead of `qt_menu.nib`
- Add plugins path for `libqcocoa.dylib`

TODO:
- [ ] Update Install documentation
- [ ] Avoid `-fPIC` for position independent code error


chris-hld:
Tested under Ubuntu 16.04
For QT5:
`sudo apt install qt5-default`

-Add fPIC flag:
Replace at ./src/Makefile.am ; line 180 and 181 with:
>if ENABLE_GUI
>AM_CPPFLAGS += \
>	-I$(srcdir)/gui \
>	-fPIC

mgeier:
Thanks @chohner, that's great!

I first tried it without the `qt5-default` packet, which was annoying since `moc` pointed to the Qt4 version. So I think it's good to recommend `qt5-default`.
If people don't want to install that, they can install `libqt5opengl5-dev` instead.
You can add to the manual, that if both Qt4 and 5 are installed, people can define this before calling `./configure`:

    export QT_SELECT=5

This `-fPIC` stuff is strange, I don't know how this can be done really platform-independently. But let's just add it, we'll see if someone complains ...

However, adding it to `AM_CPPFLAGS` doesn't make sense, since "PP" means preprocessor. I'd add it to `AM_CXXFLAGS` via `PKG_FLAGS`:

```
  PKG_FLAGS="$PKG_FLAGS $QT_CFLAGS"
  # To avoid the error "You must build your code with position independent code
  # if Qt was built with -reduce-relocations":
  PKG_FLAGS="$PKG_FLAGS -fPIC"
```

I get a few lines with this message:

    QWidget::paintEngine: Should no longer be called

I don't know if that's important.

There seems to be something wrong with the keyboard shortcuts and the context menu of the sources doesn't appear.

Did you try if the non-floating control panel works in Qt5?

umlaeute:
i'm not sure about the `-fPIC` either.
- autotools should automatically add that flag if needed, when building *libraries* (unless told so explicitely in in the `./configure` stage.
- if Qt5 requires any consumer application to be linked with `-fPIC`, it should make that explicit in it's pkg-config files (after all, how am i supposed to know whether the Qt5 library has been compiled with `-reduce-relocations` in some random distribution?)

chohner:
This seems related to QT5_widgets and people just add `-fPIC` (or `CMAKE_POSITION_INDEPENDENT_CODE` in CMAKE):
https://forum.qt.io/topic/28359/qt5-compilation-catastrophic-error/4
https://bugs.archlinux.org/task/45283
https://www.wireshark.org/lists/wireshark-bugs/201505/msg00563.html

umlaeute:
so the solution to a bug in the cmake/pkg-config files shipped with QT5 is to just add workarounds to all projects?

mgeier:
@umlaeute Yes, I think that's actually what we're supposed to do. If the Qt people would have cared about this, they would have provided a solution a long time ago.
And even if they fix it in a future version, we still need a work-around for the versions that are currently in use.

What about trying in `configure` if `-fPIC` is actually needed for successful compilation?

We could add something like this to `configure.ac` (in the `have_gui` section):

```
  AC_MSG_CHECKING([if we have to add -fPIC to make Qt happy])
  original_CPPFLAGS="$CPPFLAGS"
  CPPFLAGS="$CPPFLAGS $QT_CFLAGS"
  AC_COMPILE_IFELSE([AC_LANG_PROGRAM(
  [
    #include <QtCore/QObject>
  ], [])], AC_MSG_RESULT([no]),
  [
    AC_MSG_RESULT([yes])
    PKG_FLAGS="$PKG_FLAGS -fPIC"
  ])
  CPPFLAGS="$original_CPPFLAGS"
```

chris-hld:

chohner:
As of some days ago, qt4 was moved to the brew boneyard and `brew install qt` now installs qt5! (See https://github.com/Homebrew/homebrew-core/pull/8306)

mgeier:
@chohner What about my suggestion from https://github.com/SoundScapeRenderer/ssr/pull/58#issuecomment-261779420?

And can you please update the documentation so that @JensAhrens can include this PR in the upcoming new release?

chohner:
I think this is ready to be merged after testing this (again) on a linux system. I updated the docs and the suggested `-fPIC` stuff.

chohner:
@mgeier builds on ubunutu only if `CPPFLAGS="$CPPFLAGS $QT_CFLAGS"` is not restored (`configure.ac`, line502). Not sure why this was done?

chris-hld:
Little more info: All QT(5) related flags are missing previously listed for `CPPFLAGS`.

data/MacOSX/Makefile.am
```diff
@@ -48,22 +49,30 @@ install-exec-hook:
...
+		install_name_tool -change $(QTLIBDIR)/$$qtc.framework/Versions/5/$$qtc @executable_path/../Frameworks/$$qtc.framework/Versions/5/$$qtc $(DESTDIR)$(contentsdir)/PlugIns/platforms/libqcocoa.dylib \
+		; mkdir -p $(prefix)/Frameworks/$$qtc.framework/Versions/5 \
+		; cp $(QTLIBDIR)/$$qtc.framework/Versions/5/$$qtc $(prefix)/Frameworks/$$qtc.framework/Versions/5/ \
+		; chmod +w $(prefix)/Frameworks/$$qtc.framework/Versions/5/$$qtc \
+		; install_name_tool -id @executable_path/../Frameworks/Versions/5/$$qtc.framework/$$qtc $(prefix)/Frameworks/$$qtc.framework/Versions/5/$$qtc \
+		; install_name_tool -change $(QTLIBDIR)/QtCore.framework/Versions/5/QtCore @executable_path/../Frameworks/QtCore.framework/Versions/5/QtCore $(prefix)/Frameworks/$$qtc.framework/Versions/5/$$qtc \
+		; install_name_tool -change $(QTLIBDIR)/QtGui.framework/Versions/5/QtGui @executable_path/../Frameworks/QtGui.framework/Versions/5/QtGui $(prefix)/Frameworks/$$qtc.framework/Versions/5/$$qtc \
+		; install_name_tool -change $(QTLIBDIR)/QtWidgets.framework/Versions/5/QtWidgets @executable_path/../Frameworks/QtWidgets.framework/Versions/5/QtWidgets $(prefix)/Frameworks/$$qtc.framework/Versions/5/$$qtc \
+
```
mgeier: This empty line is not supposed to be here.

```
data/MacOSX/Makefile.am:66: error: blank line following trailing backslash
```

configure.ac
```diff
@@ -487,34 +487,35 @@ AS_IF([test x$have_gui = xyes],
 [
   # settings for Qt
   LIBS="$LIBS $QT_LIBS"
-  PKG_FLAGS="$PKG_FLAGS $QT_CFLAGS"
+  # add -fPIC flag to satisfy qt on linux
+  AC_MSG_CHECKING([Checking whether we have to add -fPIC flag to make Qt happy - ])
```
mgeier: Why the "space-dash-space" at the end?

configure.ac
```diff
@@ -487,34 +487,35 @@ AS_IF([test x$have_gui = xyes],
...
-      AC_MSG_ERROR([With Qt >= 4.7 the floating control panel is obligatory!])
-    ],
-    [test x$enable_floating_control_panel = x],
-    [
-      AC_MSG_WARN([Found Qt >= 4.7, forcing --enable-floating-control-panel!])
-      enable_floating_control_panel=yes
-    ])
-  ], [true])
+  # enable floating control panel, as it is the only option on qt5
+  enable_floating_control_panel=yes
```
mgeier: The option `--disable-floating-control-panel` doesn't work anymore (it should cause an error with Qt5).

I guess we will never use the SSR with Qt < 4.7 again, right?
Then we should get rid of the whole floating control panel option!

configure.ac
```diff
@@ -487,34 +487,35 @@ AS_IF([test x$have_gui = xyes],
 [
   # settings for Qt
   LIBS="$LIBS $QT_LIBS"
-  PKG_FLAGS="$PKG_FLAGS $QT_CFLAGS"
```
mgeier: Why did you remove this line?

This causes the error mentioned here: https://github.com/SoundScapeRenderer/ssr/pull/58#issuecomment-272497507

configure.ac
```diff
@@ -487,34 +487,35 @@ AS_IF([test x$have_gui = xyes],
...
+  # add -fPIC flag to satisfy qt on linux
```
mgeier: The following line should make the comment obsolete, since it says exactly what we are trying to do here.

mgeier:
@chohner 

> ... only if `CPPFLAGS="$CPPFLAGS $QT_CFLAGS"` is not restored (`configure.ac`, line502). Not sure why this was done?

Nothing would happen if the backup weren't restored, but I tried to keep the flags generated by the different external libraries in the `$PKG_FLAGS` variable. This way, the `$CPPFLAGS` variable stays nice and clean.

You can have a look at the contents of the variables at the end of the `./configure` run.

Actually, I've just found a "bug" there!
The Ecasound includes are added to `$CPPFLAGS`, while all other `-I` options are added (as intended) to `$PKG_FLAGS`.

If you want you can fix that!

data/MacOSX/Makefile.am
chohner: Noticed this too when building on linux and weirdly enough, the empty line is not there on OSX. Let's see what happens when removing it on linux

configure.ac
chohner: agreed.

mgeier:
@chohner Please don't merge the `master` branch back into your feature branch. This makes the history very confusing.
If you want to have the latest changes from `master`, you should rebase your branch onto `master`.

chohner:
This was githubs own "resolve conflict" in the web gui :/

mgeier:
OK, then please don't use it.

configure.ac
```diff
@@ -488,33 +488,34 @@ AS_IF([test x$have_gui = xyes],
   # settings for Qt
   LIBS="$LIBS $QT_LIBS"
   PKG_FLAGS="$PKG_FLAGS $QT_CFLAGS"
+  AC_MSG_CHECKING([Do we have to add -fPIC flag to make Qt happy])
```
mgeier: This leads to a strange sentence displayed during `./configure`:

```
checking Do we have to add -fPIC flag to make Qt happy...
```

That's why I suggested:

```
AC_MSG_CHECKING([if we have to add -fPIC to make Qt happy])
```

I must confess that I made the same mistake before, see #65.

mgeier:
I cannot open the context menu of the sources (on Debian Linux).
Each time I right-click on a source, nothing happens in the GUI and the Terminal shows:

```
QWidget::paintEngine: Should no longer be called
QWidget::paintEngine: Should no longer be called
QWidget::paintEngine: Should no longer be called
QWidget::paintEngine: Should no longer be called
```

chohner:
I can confirm `QWidget::paintEngine: Should no longer be called` on my ubuntu system. Weirdly, it works fine on OSX with only some minor glitches. I'm having a look now

chris-hld:
I'm pretty sure I've experienced the wrong indications with earlier versions, too.

chris-hld:
Fixed in https://github.com/SoundScapeRenderer/ssr/pull/73

chohner:
Ok so this is mostly ready for everyone except people with retina macbooks (or similar HDPI screens). QT5 automatically handles drawing at higher resolution (see http://blog.qt.io/blog/2013/04/25/retina-display-support-for-mac-os-ios-and-x11/) but somehow the mouse events are misplaced.

chohner:
Just tried it again with qt 5.10.1 and unfortunately nothing changed. I haven't looked at the code in a long while but simply updating didn't fix it.

chohner:
I finally found the right place to scale everything!
I will sync with @JensAhrens about some implementation details and update this PR to ask people to test :)

src/gui/qopenglplotter.h
```diff
@@ -181,6 +181,7 @@ class QOpenGLPlotter : public QGLWidget
     void _deselect_all_sources();
 
   private:
+    qreal m_scale;
```
mgeier: Just a little note on coding style: In the rest of the SSR code, non-public members start with an underscore and not with `m_`. It would be more consistent to call this `_scale`.

https://github.com/SoundScapeRenderer/ssr/blob/6cadda022e69162871f466d12dcaebd208aa9252/src/coding_style.txt#L81-L85

chris-hld:
Auto detect HDPI:
https://github.com/SoundScapeRenderer/ssr/pull/73/commits/4a95cbed7fe9e109f90c7127c7141fcd907ac2bb

chohner:
Last things to do now is to fix the positioning of the source properties (potentially in a second PR) and update the ci script

chris-hld:
Don't forget this guy:
https://github.com/SoundScapeRenderer/ssr/pull/73/commits/20ebd5b9385f57c1a2803bc55d01a006b2fd3450

chris-hld:
I think the package with the correct MOC is called `qt5-default`.

chohner:
Anyone know why is this failing with `clang: error: no such file or directory: 'gui/qclicktextlabel_moc.cpp'`?

chohner:
Very strange, on travis somehow swallows the command and instead of calling `moc` tries to call the first parameter.
Local:
```
moc -DQT_OPENGL_LIB -DQT_WIDGETS_LIB -DQT_GUI_LIB -DQT_CORE_LIB -I/usr/local/Cellar/qt/5.10.1/include/QtOpenGL -I/usr/local/Cellar/qt/5.10.1/include -I/usr/local/Cellar/qt/5.10.1/include/QtWidgets -I/usr/local/Cellar/qt/5.10.1/include -I/usr/local/Cellar/qt/5.10.1/include/QtGui -I/usr/local/Cellar/qt/5.10.1/include -I/usr/local/Cellar/qt/5.10.1/include/QtCore -I/usr/local/Cellar/qt/5.10.1/include -I./gui -o gui/qclicktextlabel_moc.cpp gui/qclicktextlabel.h
```

Travis:
```
DQT_OPENGL_LIB -DQT_WIDGETS_LIB -DQT_GUI_LIB -DQT_CORE_LIB -I/usr/local/Cellar/qt/5.10.1/include/QtOpenGL -I/usr/local/Cellar/qt/5.10.1/include -I/usr/local/Cellar/qt/5.10.1/include/QtWidgets -I/usr/local/Cellar/qt/5.10.1/include -I/usr/local/Cellar/qt/5.10.1/include/QtGui -I/usr/local/Cellar/qt/5.10.1/include -I/usr/local/Cellar/qt/5.10.1/include/QtCore -I/usr/local/Cellar/qt/5.10.1/include -I./gui -o gui/qclicktextlabel_moc.cpp gui/qclicktextlabel.h
```

chohner:
@dvzrv (Or also @mgeier)
Sorry to pull you into this unrelated PR but maybe you have some idea of whats failing on Travis' end, see https://github.com/SoundScapeRenderer/ssr/pull/58#issuecomment-387105906.

I am completely stumped on why the build fails on OSX since it works on my and @JensAhrens  machine. Especially because it fails halfway through the build process.

I also couldn't convince travis into building in debug mode, so one could inspect the state of the environment (we even tried to recreate a broken env locally where `moc` was unset - never got a similar failure like on travis).


Apart from the above, this PR is essentially ready and has been tested on macOS and linux!

dvzrv:
I'll have a look at it later today/ tonight. Not at home right now. Do you have a link to the logs somewhere?
-- 
https://sleepmap.de

chohner:
I should've updated my comment - I got it to work last night (as you can tell from the green check mark). I want to add one small fix tonight and then this PR is finally ready to merge

chohner:
@mgeier @JensAhrens @chris-hld 
This is good go to (fingers crossed).
If required I would do an interactive rebase to rewrite the history into a few concise commits before merging.

JensAhrens:
I confirm that it works beautifully on my macOS El Capitan!

The only issue that I've found when producing a bundle is incorrect references in
```
SoundScapeRenderer.app/Contents/Libraries/libvorbisenc.2.dylib
SoundScapeRenderer.app/Contents/Libraries/libvorbis.0.dylib
SoundScapeRenderer.app/Contents/Libraries/libxml2.2.dylib
```
to the local 
```
/usr/local/Cellar/libvorbis/1.3.5/lib/libvorbis.0.dylib
/usr/local/lib/libogg.0.dylib
/usr/lib/libz.1.dylib
/usr/lib/libiconv.2.dylib
```
instead of to the libs in the bundle. This issue had been around also in the Qt4 versions. Let's fix it later.

mgeier:
@chohner regarding the macOS installation on Travis-CI: probably the Qt installation isn't automatically "linked" by `brew`? Probably you should do a `brew link ...` after installation?

If this works I think it would be more elegant than adding the `bin` directory to the path.

When doing `brew install ...` you should get some information if this is a case where it is not automatically "linked".
You normally also get the relevant information for how to make `pkg-config` work.

I don't have access to a Mac, so I can't try this myself ...

chohner:
This is the additional installation info regarding the `qt`  package:
```
This formula is keg-only, which means it was not symlinked into /usr/local,
because Qt 5 has CMake issues when linked.

If you need to have this software first in your PATH run:
  echo 'export PATH="/usr/local/opt/qt/bin:$PATH"' >> ~/.zshrc

For compilers to find this software you may need to set:
    LDFLAGS:  -L/usr/local/opt/qt/lib
    CPPFLAGS: -I/usr/local/opt/qt/include
For pkg-config to find this software you may need to set:
    PKG_CONFIG_PATH: /usr/local/opt/qt/lib/pkgconfig
```

Brew doesn't allow you to link qt unless you use `brew link qt --force`. I simply wrongly assumed that `PKG_CONFIG_PATH` is enough but the system just never knew about the bin folder (and the `moc` executable inside). I think for now we should leave it like this as this is probably the most common way to have qt set up on a Mac.

mgeier:
OK, thanks, since `brew` suggests manipulating the `$PATH` variable, I guess it's fine if we use it in the Travis-CI settings.

But we should also add it to the documentation, right?
Or is it already there?

It looks like for Qt5, the `moc_location` information doesn't exist anymore (at least on my Debian Linux):

    pkg-config --variable=moc_location Qt5Core

(I guess that was the reason why the `moc` command disappeared in https://github.com/SoundScapeRenderer/ssr/pull/58#issuecomment-387105906)

The only thing that seems to be available is `host_bins`, but that's of limited usefulness:

    pkg-config --variable=host_bins Qt5Core

I think at some point we should remove the `moc_location` thing from `configure.ac` and I guess there isn't a real replacement we could use instead ... or is there?

chohner:
I think for now we can expect people to correctly install the listed dependencies. Reworking the `configure.ac` (or maybe even more of the build system, looking at `CMake`) is a great idea but also outside of the scope of this PR, already big enough as it is.

chris-hld:
Yes!
PLEASE merge!!

JensAhrens:
Sorry, I've lost track here. Are there any more requested changes?

chris-hld:
Nope, it's more than enough for one commit ;)

mgeier:
Thanks to all who were working on this!

umlaeute:
does this warrant a new release?

PR #59, opened 2016-11-18
Replaced boost by c++11 and asio

chohner:
This completely strips Boost from the network interface:
- Include asio.hpp and define ASIO_STANDALONE
- boost > std: (enabled_shared_from_this, shared_ptr, bind,
asio::placeholders, system::, thread)
- boost::asio > asio (ip::tcp::socket, ip::tcp::acceptor,
ip::tcp::endpoint, io_service, error_code, streambuf, async_write)
- boost::asio::deadline_timer becomes asio::steady_timer (to work with
std::chrono)
- boost::posix_time > std::chrono

This *should* compile already after installing asio (`brew install asio` on OSX, `apt-get install libasio-dev` on unix).

Todo:
- [x] Adjust build environment (check against asio instead of boost)
- [x] Update readme on installing asio and enforcing c++11


chris-hld:
Yep, compiles under Ubuntu 16.04.

mgeier:
Cool, that works great for me on Debian testing!

All the ugly Boost stuff in `configure.ac` can now be replaced with those few lines:

```
ENABLE_FORCED([ip-interface], [network (TCP/IP) interface],
[
  CPPFLAGS_backup="$CPPFLAGS"
  CPPFLAGS="$CPPFLAGS -DASIO_STANDALONE"
  AC_CHECK_HEADER([asio.hpp], , [have_ip_interface=no])
  CPPFLAGS="$CPPFLAGS_backup"
])
```

I think it makes sense to define `ASIO_STANDALONE` in the code where you did it, therefore it should be un-defined again after checking for `asio.hpp`.

Or probably it would be even better to do something like this (and removing the `CPPFLAGS_backup` lines above):

```c++
#if !defined(ASIO_STANDALONE)
#define ASIO_STANDALONE
#endif

File: doc/manual/operation.rst [hunk 1]
```diff
@@ -1312,17 +1318,6 @@ This happens when two or more instances of the SSR are started with the IP serve
 Start all (or at least all instances higher than 1) with the ``-I`` flag to disable the
 IP interface. 
 
-IP interface isn't selected although boost libraries are installed
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
```
mgeier: Please don't remove this. Instead, add "This issue was resolved in ...".

umlaeute:
maybe the use of `ASIO_STANDALONE` should be made settable by the person who is compiling (i don't have any special use-case in mind though)

chohner:
> maybe the use of ASIO_STANDALONE should be made settable by the person who is compiling (i don't have any special use-case in mind though).
That switch would then add boost dependencies deep inside the asio library that you would need to tell the compiler/linker about and so on.

umlaeute:
are these dependencies build-deps or runtime dependencies?

why does asio provide a non-standalone option (and even defaults to that) if it is not desirable to use it?

chohner:
> are these dependencies build-deps or runtime dependencies?

Build time. But some problems we were seeing on OS X also disappeared on this build so I don't think there's any reason to allow building with boost.

>why does asio provide a non-standalone option (and even defaults to that) if it is not desirable to use it?

I think only for historical reasons. The standalone option is still (comparatively) new and relies on c11. Since afaik there's plenty of c11 code throughout SSR, the only upside of building with boost (building with on std<c11) disappears.

chohner:
Small update from my side: I can confirm the interface to be working with the SSR android app.

doc/manual/operation.rst
mgeier: Please add the appropriate SSR version number.

mgeier: @chohner When you are done, can you please squash your commits?

JensAhrens: Yes, 0.5.0 is it!

JensAhrens: ... and I'm happy to include this PR.

chohner:
Added a line on being fixed in version `0.5.0` and clarified the second line.  The squashing can/should be done on your part when merging ([squash&merge is available directly in the github GUI](https://github.com/blog/2141-squash-your-commits)). 
Feel free to use something like this as the commit description:
```
* replaced boost::asio by c++11 and asio
* cleaned up configure.ac
* rename src/boostnetwork to src/network
* update manual
```

mgeier:
Thanks @chohner.

You squashing your commits (or somehow else pushing them into shape) places the burden of coming up with commit messages on you, that's all.

Anyway, @JensAhrens will merge it, so I don't really care.

chris-hld:

mgeier:
Apparently, I've approved this too quickly. Creating the manual produces a few errors now.

JensAhrens:
I fixed the manual.

PR #63, opened 2016-11-30
Reworked README to contain useful information.

dennisguse:
COPYING was renamed, so the license shows in Github.
INSTALL was removed as it does not contain *any* useful information (just a generic file).


umlaeute:
:-1: 

what's wrong with generic `INSTALL` information? not that specific information might not be better; but if you are *not* used to `./configure && make && make install`, then generic instructions can be helpful as well.
after your commit, there is *no* information, which i don't think is any better.

also, what is wrong with a separate `AUTHORS` file?


finally: as of now, your PR actually results in a **failed build**, because the build-system *requires* the files `README`, `INSTALL`, `COPYING` and  `AUTHORS` to be present.
i think this is substantially worse than not having the LICENSE show up on the github page.

hagenw:
You have spelled the SoundScape Renderer wrongly (SoundScapeRenderer) in the new README.

dennisguse:
INSTALL: What do you think about about a link to the INSTALL file and mentioning `autoconf && ./configure && make && make install` in the README.
I think, if you are not used to this command, you will most probably not be able to understand the INSTALL anyhow ;)

AUTHORS: the file suggests that only these persons worked on the SSR.
However, even in the GIT history there are several not mentioned people: https://github.com/SoundScapeRenderer/ssr/graphs/contributors
So, who is going to decide who will be named?

In addition, I have the question if it is most important to name all people everywhere (e.g., SSR_AUTHORS, SSR_AUTHORS_QT used in the code.)?

@hagenw: fixed name.

mgeier:
@dennisguse Thanks for this PR, @umlaeute @hagenw Thanks for the justified criticism. I agree.

re `INSTALL`:

Please leave it as is, except if there is a newer version you should of course update it.

Did you actually read this file?
It is actually very helpful to understand how `configure` et al. are supposed to be used.
If is surprisingly informative, if people would only care to read it ...

If some of the information in there isn't appropriate, it doesn't mean that it is wrong, rather the SSR build system should be changed to fit the description.
If you find something like this, please create a PR.

re `AUTHORS`:

Removing this file breaks a lot of things. This file is the sole source for a list of authors, which is used in the manual, in the `--version` output, in the "about" page of the GUI (and probably somewhere else which I forgot).

> if it is most important to name all people everywhere

Sure it is!
Why would you want to hide the authors?

The only place where I think it is not appropriate to mention them, is in the `README`.

> there are several not mentioned people

There are two.

You are right, @umlaeute should be mentioned there, not merely because of his 1 commit, but because of the Debian package which he created (and maintains!) and because of his continuous support here on Github.

Please add him to `AUTHORS` instead of removing the file. You can make a separate PR, since this is not related to the README.

The contribution of @c-bebop is also appreciated, but I think the amount of her/his contributions is still below the threshold of adding her/him to the `AUTHORS`.
If anyone disagrees, please speak up!

re `README`:

I'll comment on this once the other stuff is out of the way.

dennisguse:
Addressed the issues: no files are removed now.

mgeier:
@dennisguse Thanks for the update!

Did you intentionally leave the old `README` in place?
Did you intentionally duplicate the author information from `AUTHOR`?
If yes, why?

dennisguse:
Removed old README.
About `AUTHORS` Yes, I duplicated these on purpose.
If you think the authors should not be named in the `README.md`, then please remove them.
On the long run I would, however, prefer removing `AUTHORS`.

PS: What are the rules for becoming named as an author?

umlaeute:
i'm pretty sure that a file called `README` (without a suffix) is necessary for autotools.
thus you shouldn't completely remove it.
either use a symlink (which could lead to confusion on filesystems that don't support symlinks), or use it to refer the people to the proper file.

```
$ cat README
Please see README.md
$
```

#### regarding *How to become an AUTHOR?*

make a substantial contribution to the code?
btw, I don't think that *I* should go into that list.

mgeier:
I looked around a bit:

* Wikipedia mentions some common patterns: https://en.wikipedia.org/wiki/README

* The GNU Coding Standards say there *should* be a file named `README` and they explain what should be in there https://www.gnu.org/prep/standards/standards.html#Releases

* Using Automake, you can choose different levels of "strictness":
  * https://www.gnu.org/software/automake/manual/automake.html#Strictness
  * https://autotools.io/automake/options.html#automake.options.flavors

We don't have to adhere to the GNU Coding Standards, and actually we don't, since we use the `foreign` option in `configure.ac` and we don't have a `ChangeLog` file, for example.
However, unless we have a good reason to depart from the GNU standards, it doesn't hurt to follow them.

@dennisguse 

Thanks for the update.

> About `AUTHORS` Yes, I duplicated these on purpose.

I'd like to avoid duplication, unless there is a good reason to duplicate.
Is there?

> If you think the authors should not be named in the `README.md`, then please remove them.

I don't think they should be named there.
Why do you think they should be named there?

And no, I won't remove them. That's not how code review works.
Maybe there is a good reason to put them there, we can discuss that here.
And if there's not, you shouldn't include them in the PR in the first place.

> On the long run I would, however, prefer removing `AUTHORS`.

Why do you hate that file so much?
And what would you replace it with?

PR #70, opened 2017-03-08
src/*: Removing vim:fdm (foldmethod) settings from all source files.

dvzrv:
Removing restrictive vim foldmethod settings from all source files.

### DIFF
--- a/apf/apf/biquad.h
+++ b/apf/apf/biquad.h
@@ -347,4 +347,3 @@ SosCoefficients<T> bilinear(LaplaceCoefficients<T> coeffs_in, int fs, int fp)
 
 // Settings for Vim (http://www.vim.org/), please do not remove:
 // vim:softtabstop=2:shiftwidth=2:expandtab:textwidth=80:cindent
-// vim:fdm=expr:foldexpr=getline(v\:lnum)=~'/\\*\\*'&&getline(v\:lnum)!~'\\*\\*/'?'a1'\:getline(v\:lnum)=~'\\*\\*/'&&getline(v\:lnum)!~'/\\*\\*'?'s1'\:'='
--- a/apf/apf/blockdelayline.h
+++ b/apf/apf/blockdelayline.h
@@ -318,4 +318,3 @@ class NonCausalBlockDelayLine : private BlockDelayLine<T, Container>
 
 // Settings for Vim (http://www.vim.org/), please do not remove:
 // vim:softtabstop=2:shiftwidth=2:expandtab:textwidth=80:cindent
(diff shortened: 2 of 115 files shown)


mgeier:
Thanks! You are right, those shouldn't be in there.

For completeness, you could also remove them from the `apf/` files?

dvzrv:
Sure thing!
btw: will apf be included as a git submodule in the future?

dvzrv:
I sqashed the last commit and renamed the message.

mgeier:
> will apf be included as a git submodule in the future?

I don't know, should it?

Initially, I decided for `git subtree` instead, because then I'm the only one who has to worry about it. If you are interested, have a look at [update_apf.txt](../blob/master/update_apf.txt).

Using `git submodule` can be quite confusing for beginner and intermediate level Git users. I know it took me quite a while to get used to it ...

chohner:
I think the idea of `git submodule` is both pretty elegant and straightforward to understand & maintain. But you are correct, you do need to run two more commands when cloning initially. I'd still argue that it's worth it.

dvzrv:
`git submodule` seems indeed to be the more elegant solution compared to `git subtree`.
So far I found submodules quite easy to work with and it's a very well documented feature, too. It would most likely make testing of udpated apf versions a bit easier.  

Ultimately as maintainer you'll be responsible for the "current version" though.

mgeier:
Regarding `submodule` vs. `subtree`:

I think it really depends on how the repository is being used. And this might change over time.

If the APF were a separately developed project with actual users and actual PRs created for it, it would probably make more sense to make it a `submodule`.

However, since it is currently mainly developed from inside of the SSR repo and the flow of information is mainly from SSR to APF, I think `subtree` actually makes more sense.

I guess there is one use case that is crucial:
If somebody wants to make a PR that changes things in both the SSR and the APF sources, this is only possible with `subtree`.
With `submodule`, one would have to make a separate PR for APF and then another one for SSR, changing the submodule to point to the APF PR. After the APF PR is merged, the SSR PR has to be changed to point the submodule back to APF `master` (unless it was fast-forward merged), and then the SSR PR can be merged.

Granted, there haven't been many PRs that manipulate both the APF and SSR code base, but I think it's nonetheless better if contributors don't have to worry about that.

What are the disadvantages of the current `git subtree` based approach?

> It would most likely make testing of udpated apf versions a bit easier.

Sure, but only if there was such a thing as "updated APF versions".
Currently, the APF doesn't change "on its own".

If at some point in the future this starts happening, we can switch to `git submodule`.

dvzrv:
> Currently, the APF doesn't change "on its own".
> 
> If at some point in the future this starts happening, we can switch to git submodule.

Sounds reasonable!

Feel free to merge (so I can stop cherrypicking ;-) ).

mgeier:
@JensAhrens Can you please merge whenever it's convenient for you?

umlaeute:
just another note on `git submodules`: github doesn't include submodules in the auto-generated tarballs of tagged releases

so using submodules can make the automatic tarballs much less useful.

PR #73, opened 2017-03-17
Source properties are Qt5 ready

chris-hld:
This is a follow up to https://github.com/SoundScapeRenderer/ssr/pull/58.
The property box didn't reflect the mute and fixed source status.
This is because mouse click events were not passed to QCheckBox.

### DIFF
--- a/src/gui/qsourceproperties.cpp
+++ b/src/gui/qsourceproperties.cpp
@@ -27,8 +27,8 @@
 /// @file
 /// TODO: add description
 
-#include <QPalette> 
-#include <QSizePolicy>
+#include <QtGui/QPalette> 
+#include <QtWidgets/QSizePolicy>
 
 #include "qsourceproperties.h"
 
--- a/src/gui/qsourceproperties.cpp
+++ b/src/gui/qsourceproperties.cpp
@@ -343,7 +343,7 @@ void QSourceProperties::update_displays(const Source& source,
   {
     _properties_display->setText(source.properties_file.c_str());
   }
-  this->raise();
+
 }
 
 void QSourceProperties::_expand()
--- a/src/gui/qsourceproperties.cpp
+++ b/src/gui/qsourceproperties.cpp
@@ -43,11 +43,7 @@ using namespace apf::math;
 //#define ROWHEIGHT 40
 
 QSourceProperties::QSourceProperties(QWidget* parent)
-#ifdef __APPLE__
 : QFrame(parent, Qt::Window | Qt::CustomizeWindowHint),
-#else
-    : QFrame(parent),
-#endif
     _create_new_source(false)
 {
   this->setAutoFillBackground(true);
--- a/src/gui/qsourceproperties.cpp
+++ b/src/gui/qsourceproperties.cpp
@@ -343,7 +343,7 @@ void QSourceProperties::update_displays(const Source& source,
   {
     _properties_display->setText(source.properties_file.c_str());
   }
-
+  this->raise();
 }
 
 void QSourceProperties::_expand()
--- a/src/gui/qsourceproperties.cpp
+++ b/src/gui/qsourceproperties.cpp
@@ -43,11 +43,7 @@ using namespace apf::math;
 //#define ROWHEIGHT 40
 
 QSourceProperties::QSourceProperties(QWidget* parent)
-#ifdef __APPLE__
 : QFrame(parent, Qt::Window | Qt::CustomizeWindowHint),
-#else
-    : QFrame(parent),
-#endif
     _create_new_source(false)
 {
   this->setAutoFillBackground(true);
@@ -347,7 +343,7 @@ void QSourceProperties::update_displays(const Source& source,
   {
     _properties_display->setText(source.properties_file.c_str());
   }
-
+this->raise();
 }
 
 void QSourceProperties::_expand()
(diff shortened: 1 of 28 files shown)


src/gui/qsourceproperties.cpp
```diff
@@ -413,20 +413,21 @@ void QSourceProperties::_set_doppler(bool flag)
...
+//       e->type() == QEvent::MouseButtonPress ||
+//       e->type() == QEvent::MouseButtonRelease ||
+//       e->type() == QEvent::MouseMove)
+//   {
+//     e->accept();
+//     return true;
+//   }
+//   else return false;
+//
+// }
```
chris-hld: Is there any reason for this special catching?
We could remove it otherwise...

chohner:
I think all trailing spaces were deleted in your last commit which makes it hard a little hard to follow. Did you just comment out `bool QSourceProperties::event(QEvent *e)` and `virtual bool event(QEvent *e);`?

chris-hld:
Yes. I haven't found a way to explicitly pass the mouse events to the checkbox, but I think it isn't necessary. We should remove this.

chohner:
Cool, this is working on my OSX machine - numbers and checkboxes etc are populated/updated correctly!

Just one last quirk - the source dialog is drawn BEHIND the background when I want to drag the source with a left click while having the dialog open. Is that also happening on unix?

chris-hld:
No, I didn't notice. But that might be due to the fact, that the properties box appeared on top of the source, not beside. There was a special case for mac defined. I updated this in `quserinterface.cpp `

chohner:
again a very unreadable commit due to all the deleted whitespace :/ Had to look 3 times until I found that you changed line `1147`.

Could you disable that in your editor? While its probably not bad practice to eliminate trailing whitespace, this should probably be its own PR and agreed upon by @mgeier.

mgeier:
Great work Chris & Chris!
I tested it on Linux and it works nicely.
What's missing from #58 is removing everything about the "floating control panel", anything else?

I agree that large-scale whitespace changes should be done separately, if possible.
Those cases of trailing whitespace should have never been committed in the first place, but that's another story.

I also don't really like the shape of the history. Ideally it should be based in a single line of descent (and whenever necessary re-based) on `master`, and maybe some commits could be combined to be more meaningful.

This time I'm willing to let that slip, but in the future you should keep that in mind.

chohner:
I'm happy to incorporate the most recent changes in my PR (without the whitespace stuff), clean up the history and remove the floating panel stuff tomorrow.
Maybe we can get this ready tomorrow and then @JensAhrens can look into finally making a new release?

chris-hld:
You are right about the whitespace stuff, should have handled that more carefully.
What's the best practice for adding on top of a pull request? Rebase on the remote the pull request is coming from?

mgeier:
@chris-hld

> What's the best practice for adding on top of a pull request?

I'm not sure about best practices, but I would suggest this:

If you have write access, you can simply add commits to the existing PR. But I guess you should ask the creator of the PR if that's OK.

If you don't have write access (which I assume is your case) or if you just don't want to interfere too much, you can make a separate branch (and a separate PR) based on the original PR (as you did).

The original pull requester is now free to cherry-pick your commits into his own PR.
And you are free to re-base or squash or delete the commits from the original PR that are now in your branch/PR.
And if either one of the PRs changes, the other one can also apply those changes or parts of them or just ignore them.

In our case, I guess we have to decide at some point if we should merge #58 or #73.
Or @JensAhrens just cherry-picks some commits from both PRs into `master` and closes both PRs.

As always with Git, there are many possible work flows ...

Since @chohner already offered to make changes to his PR #58, I'd suggest that he does that, and *afterwards* he can just re-base your (@chris-hld's) commits on top of his stuff. Before he does that, you (@chris-hld) should make the changes regarding whitespace (and any other changes you want to make) in your PR #73.

If it turns out that @chris-hld's changes don't apply cleanly on @chohner's new changes, then it's up to @chris-hld to make the necessary changes on his PR.

Whenever you are finished and you are both happy with the result, just close one of the PRs.

Does that sound like a plan?

chohner:
@chris-hld When you're cleaning up your commits, could you delete the `QEvent` rather than commenting them out?

chris-hld:
Ready for cherry-picking.
 	eaf631f and  c680b99 .
Can be squashed if preferred.

JensAhrens:
It all looks fine on my external display. But when I use the built-in display of my MacBook, then the scene window opens in a funny zoom state. It all looks normal as soon as I double-click on the background to re-set the zoom. However, the mouse events occur at wrong positions.

chohner:
Oh no, I remember that too. Should we continue the discussion in https://github.com/SoundScapeRenderer/ssr/pull/58? I moved the commits over and just cleaned up some FLOATING window stuff. There are several points where window position etc is touched, hopefully this can be fixed easily. Might be something weird between OpenGL and retina displays?

chris-hld:
Does this help:

> set the system environment variable:
>QT_AUTO_SCREEN_SCALE_FACTOR=1

http://stackoverflow.com/questions/24367355/automatic-rescaling-of-an-application-on-high-dpi-windows-platform
http://doc.qt.io/qt-5/highdpi.html

Seems like a good idea anyhow.

chohner:
Good find, somehow `QT_AUTO_SCREEN_SCALE_FACTOR` slipped passed me. I already started messing around with `devicePixelRatio()` inside the painting code but maybe QT can handle this more gracefully?

chris-hld:
@JensAhrens Does this fix the window zoom state?

chohner:
It doesn't unfortunately. All "simple" fixes mentioned there (like `Qt::AA_EnableHighDpiScaling`) don't really change anything.

I'm still playing around with it but currently the only way to make some stuff work was for me to remap the location of the mouse click based on the `devicePixelRatio()`, like this in  `qopenglplotter.cpp, line 951 (ssr::QOpenGLPlotter::_get_openGL_pos)`, which fixes the lasso :
```
  qreal pixelRatio = devicePixelRatio();

  win_x = static_cast<GLfloat>(x)*pixelRatio;
  win_y = static_cast<GLfloat>(viewport[3]) - static_cast<GLfloat>(y)*pixelRatio;
```

JensAhrens:
Same here, no improvement with `QT_AUTO_SCREEN_SCALE_FACTOR=1`.

chris-hld:
There were a lot of bug fixes in latest QT5 releases. Many of them tackled high DPI screens.
Maybe you should update QT5 and try again?

chris-hld:
Moved over to https://github.com/SoundScapeRenderer/ssr/pull/58.

PR #74, opened 2017-03-20
Toggle Solo state by key "s"

chris-hld:
The key "s" did not toggle the solo state, as you would expect it from all other multitrack audio softwares I'm familiar with. Very inconvenient, especially for demos, as you have to unselect all sources first and then hit "s" again.
The key "m" already shows the expected behaviour for mute (toggle) state.

- This PR improves the shortcut "s" and matches it to shortcut "m"

### DIFF
--- a/src/gui/qgui.cpp
+++ b/src/gui/qgui.cpp
@@ -40,7 +40,7 @@ QString qt_style_sheet =
 #else
   "QWidget {                                               \n"
   "   color: black;                                        \n"
-  "   font: normal 9pt \"Monospace\";                      \n"
+  "   font: normal 10pt \"Monospace\";                     \n"
   "   }                                                    \n"
 #endif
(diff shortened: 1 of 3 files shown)


chris-hld:
Is there any benefit in not renaming a nested iterator?

Also, I want to propose darkening the colour of muted sources. 
3b6531f755fc60502bdf911f8c243137db899a79
This makes it so much easier to recognize whether sources are in mute state.

JensAhrens:
The darkening seems like a good idea. 

I have no opinion on the iterator...

chris-hld:
I will split this PR into two.

chris-hld:
https://github.com/SoundScapeRenderer/ssr/pull/89
https://github.com/SoundScapeRenderer/ssr/pull/90

Issue #67, opened 2017-02-12
Can't build this with latest Qt...


Harvie:
I can't build this with latest Qt...

```
g++ -DHAVE_CONFIG_H -I.  -I./../apf  -DSSR_DATA_DIR=\"/usr/local/share/ssr\" -D_POSIX_PTHREAD_SEMANTICS  -I./gui -I/usr/include/libecasoundc -Wall -Wextra -pedantic -Wno-long-long -Winit-self -Wcast-align -Wmissing-declarations -Wredundant-decls -Woverloaded-virtual -Wnon-virtual-dtor -Wwrite-strings -Wstrict-overflow=0 -I/usr/include/libxml2 -DQT_SHARED -I/usr/include/qt4 -I/usr/include/qt4/QtOpenGL -I/usr/include/qt4 -I/usr/include/qt4/QtGui -I/usr/include/qt4 -I/usr/include/qt4/QtCore -fomit-frame-pointer -ffast-math -funroll-loops -march=k8 -mmmx -msse -mfpmath=sse -msse2 -mfpmath=sse -DNDEBUG -g -std=c++11 -pthread -O3 -MT gui/qclicktextlabel_moc.o -MD -MP -MF $depbase.Tpo -c -o gui/qclicktextlabel_moc.o gui/qclicktextlabel_moc.cpp &&\
mv -f $depbase.Tpo $depbase.Po
gui/qclicktextlabel_moc.cpp:15:2: error: #error "This file was generated using the moc from 5.8.0. It"
 #error "This file was generated using the moc from 5.8.0. It"
  ^~~~~
gui/qclicktextlabel_moc.cpp:16:2: error: #error "cannot be used with the include files from this version of Qt."
 #error "cannot be used with the include files from this version of Qt."
  ^~~~~
gui/qclicktextlabel_moc.cpp:17:2: error: #error "(The moc has changed too much.)"
 #error "(The moc has changed too much.)"
  ^~~~~
gui/qclicktextlabel_moc.cpp:21:1: error: ‘QT_WARNING_PUSH’ does not name a type
 QT_WARNING_PUSH
 ^~~~~~~~~~~~~~~
gui/qclicktextlabel_moc.cpp:26:2: warning: extra ‘;’ [-Wpedantic]
 };
  ^
gui/qclicktextlabel_moc.cpp:32:14: error: ‘qt_meta_stringdata_QClickTextLabel_t’ does not name a type
[...]

Issue #64, opened 2016-11-30
Remove cleanse.sh


dennisguse:
cleanse.sh: `# Shell-Script for removing all files that don't belong into the Git repo. `

This can be easily done using `git clean -f -d`.
This command _really_ removes everything that is not part of the GIT.

mgeier:
Probably the comment is misleading, but `cleanse.sh` is supposed to only remove those files that are created by the build system, not just all files that are not managed by Git.

I might have some test files lying around, probably some text files with notes, whatever. I normally don't want to remove those.

Feel free to submit a PR that fixes the aforementioned comment.

PR #75, opened 2017-03-30
.gitignore: Adding entries for local build, ctags and compile_commands.

dvzrv:
I would like to update the entries for .gitignore.
The added files are all files, that are being generated when building the project, except _compile_commands.json_ (which is generated by [bear](https://github.com/rizsotto/Bear) and can be used in conjunction with [YouCompleteMe](https://github.com/Valloric/YouCompleteMe) to have nice completion in the case, where a _.h_ and a _.cpp_ file are present), _tags_ (used for ctags) and _ssr-qt4.patch_, which is my personal patch to fix _configure_ to build with the outdated moc ([#67](https://github.com/SoundScapeRenderer/ssr/issues/67)).
None of them should be added to the repository (afaik) and having them in .gitignore helps a great deal in keeping track of what is actually important when doing a `git status`.

To get rid of all of them it's always possible to do a `git clean -fxd`, in case that's needed.


mgeier:
I don't know, this doesn't sound right ...

Did you try a VPATH build?

https://www.gnu.org/software/automake/manual/html_node/VPATH-Builds.html

This would break our little helper script `data/local_ssr.sh`, but probably this could be fixed or completely replaced by some other workflow.

Either way, it would also be nice if those build details would be documented somewhere, or is it?

This should already get rid of most of your cases.

IMHO it doesn't make sense to add stuff that's specific to your environment.
You can do that in your own global ignore file.

If the YCM thing is helpful to others, we can upgrade it to an official recommendation and describe it in the manual.
BTW, I'm also using YCM (which is great), but with a manually written `.ycm_extra_conf.py`.

Another thing we can probably clean up, is the `cleanse.sh` script, see #64.

dvzrv:
> Did you try a VPATH build?

No, I tried it the way it is documented. ;-)
If VPATH is the preferred way, it should go to the manual, I guess.
Speaking of which: Working on a refactoring, to add/update build and install instructions ([not done yet](https://github.com/dvzrv/ssr/tree/installation_documentation)).
It surely is a little less straight forward (as `autogen.sh` can't be called from `build/`), but after all doesn't make much of a difference.

After using VPATH, there would still be the files generated from `autogen.sh`.

```
        Makefile.in
        aclocal.m4
        autom4te.cache/
        autotools/
        build/
        configure
        configure.orig
        data/MacOSX/Makefile.in
        data/Makefile.in
        man/Makefile.in
        src/Makefile.in
        src/config.h.in

```

> This would break our little helper script data/local_ssr.sh, but probably this could be fixed or completely replaced by some other workflow.

Well, it can stay as it is, if the files are added to .gitignore. No harm in doing that. :-)

If VPATH is/ will be the preferred way, can't the symlink generation for `data/` just be adapted for it during `make` and a copy of `data/{ssr,local_ssr.sh}` be placed in `build/data/` or something similar?
That way only the above mentioned files (and other "specials", such as ctags, etc.) could be added to .gitignore.

The reason why I'd like to add the files is, that it is getting very crowded in `git status`, once you used make. It's just easier to see, what you're working on, without all those files, that should never be added to the repository anyways.

> IMHO it doesn't make sense to add stuff that's specific to your environment.

You are right. Will remove the patch file.

> You can do that in your own global ignore file.

Unfortunately it's not that easy (without renaming) in the case of the patch file, as git is used in all of the [AUR](https://aur.archlinux.org)'s packages. Ignoring globally would effect my package maintaining.
However, I can live with a single untracked file in a repository ;-)

> If the YCM thing is helpful to others, we can upgrade it to an official recommendation and describe it in the manual.

That sounds like a plan!

> BTW, I'm also using YCM (which is great), but with a manually written .ycm_extra_conf.py

Oh  nice! Could I have a peak? I have opted for bear, as it seemed the faster solution at that time.

> Another thing we can probably clean up, is the cleanse.sh script, see #64.

Ah, I see. Well, if that script (and VPATH) is to stay, how about adding `build/` to the things to be removed and fixing that comment?
Having `cleanse.sh` does not conflict with having the above mentioned files in `.gitignore` btw. as it offers the less drastic alternative to `git clean -fxd`.

mgeier:
The smaller list you mentioned sounds much more reasonable to me.
It would be interesting how other autotools-based projects are handling this.

I wouldn't add `build/` itself though, since this is just an example and users can use whatever directory they want. It doesn't even have to be inside of the main SSR directory.

> If VPATH is the preferred way, it should go to the manual, I guess.

I don't know if it is, I just thought about it when I wrote my previous comment, I haven't actually used it that much.
I have used `./cleanse.sh` quite regularly, whenever I wanted to keep an eye on the `git status`, but I admit that that's not the best way to do things.
I could imagine using VPATH builds in the future.

BTW, improvements of the manual are highly appreciated!

> If VPATH is/ will be the preferred way, can't the symlink generation for `data/` just be adapted for it during `make` and a copy of `data/{ssr,local_ssr.sh}` be placed in `build/data/` or something similar?

I'm sure this could be fixed, I myself am just not very eager to do that in the moment.

> The reason why I'd like to add the files is, that it is getting very crowded in `git status`, once you used make. It's just easier to see, what you're working on, without all those files, that should never be added to the repository anyways.

I understand. I'm just not sure if adding a gazillion of files to `.gitignore` is the right solution.
But probably it is ... I don't know, I'm just asking ...

> > BTW, I'm also using YCM (which is great), but with a manually written `.ycm_extra_conf.py`
>
> Oh  nice! Could I have a peak?

Sure:

```python
flags = [
    '-x', 'c++',
    '-std=c++11',
    '-Isrc',
    '-Iapf',
    '-Isrc/boostnetwork',
    '-Isrc/gui',
    '-I/usr/include/libecasoundc',
    '-I/usr/include/libxml2',
    '-I/usr/include/qt4',
    '-I/usr/include/qt4/QtGui',
    '-I/usr/include/qt4/QtCore',
    '-I/usr/include/qt4/QtOpenGL',
    '-include', 'apf/pointer_policy.h',
    '-include', 'apf/dummy_thread_policy.h',
    '-D_REENTRANT',
    '-DENABLE_AAP',
    '-DENABLE_BINAURAL',
    '-DENABLE_BRS',
    '-DENABLE_GENERIC',
    '-DENABLE_NFC_HOA',
    '-DENABLE_VBAP',
    '-DENABLE_WFS',
    '-DENABLE_GUI',
    '-DENABLE_IP_INTERFACE',
    '-DENABLE_ECASOUND',
    '-DENABLE_INTERSENSE',
    '-DENABLE_POLHEMUS',
    '-DENABLE_RAZOR',
    '-DENABLE_VRPN',
    '-DPACKAGE_STRING="dummy"',
    '-DPACKAGE_URL="dummy"',
    '-DPACKAGE_BUGREPORT="dummy"',
    '-DSSR_COPYRIGHT="dummy"',
    '-DSSR_AUTHORS="dummy"',
    '-DSSR_DATA_DIR="dummy"',
]

def FlagsForFile(filename):
    return {'flags': flags, 'do_cache': True}
```

I haven't committed it because it contains a few platform-specific paths.
But now that I'm thinking about it, this could probably created by the build system?

> I have opted for bear, as it seemed the faster solution at that time.

I've never tried it. Is it worth trying?

dvzrv:
> The smaller list you mentioned sounds much more reasonable to me.
> It would be interesting how other autotools-based projects are handling this.

See the updated `.gitignore`.
It still features the tags file (for ctags) and `compile_flags.json` (for ycm) though, which I hope is okay.
The latter could probably also be left out in the case of "VPATH is teh future".

> I could imagine using VPATH builds in the future.

Okay, let me rephrase then: Should it be the recommended way in the future? :)

> BTW, improvements of the manual are highly appreciated!

Will do a pull request for it as soon as I have some more of the network stuff figured out and have had a second look over it.

> I haven't committed it because it contains a few platform-specific paths.
> But now that I'm thinking about it, this could probably created by the build system?

Not sure. That would require make before being able to work on it with vim + ycm.
I would really like it to be included though (and thanks for the sample!).
Hmm, is there any way of getting the paths to global includes platform-independently?

> I've never tried it. Is it worth trying?

Well, it "only" requires a `bear make` instead of `make` (that in return currently requires the qt4 patch to be applied to `configure` for me), which generates the `compile_commands.json`.

Both `.ycm_extra_conf.py` and `compile_commands.json` are searched for upwards the tree from the currently opened file (and then recursively downwards, as far as I understand it) and then applied/called once, according to the [ycm manual](https://github.com/Valloric/YouCompleteMe/#c-family-semantic-completion). In any case: Both solutions have the hen-egg problem (kind of), as you would have to call `make` before being able to use them (platform-specific).

A good solution would probably be to include several versions of the `.ycm_extra_conf.py` somewhere in the documentation and add them to .gitignore, as they'll (most likely always) be platform specific.
While it only makes sense to ignore `compile_commands.json`, but maybe include a hint on how to use bear to generate it (if it should be supported at all).

umlaeute:
personally i don't think that `VPATH` should be recommended (i just don't see a reason to recommend it).
it should be fully supported though (and afaik it already is) - at least *I* need it (with my Debian hat on)

mgeier:
I think for "normal" users, we shouldn't recommend VPATH builds. But I think we could describe it as an alternative in the (not yet existing) `CONTRIBUTING.rst` file. For now we have only this very short section on installing from the Git repo: http://ssr.readthedocs.io/en/latest/operation.html#compiling-and-installing-from-the-git-repository. This could be extended with some additional information for contributors.

The main reason for using VPATH builds would be to avoid pollution of the source directory, but it looks like we can't avoid that completely because `./autogen.sh` will do some pollution anyway.
Another *theoretical* reason would be to switch between building different variants of the SSR, e.g. one with GUI and one without GUI, but I don't know if anyone would like to do that ...

@umlaeute Do you think it makes more sense to put all the compiler-generated files into `.gitignore`?

@dvzrv 

> That would require `make` before being able to work on it with vim + ycm.

Only `./configure` is needed.
And at some point you'll run this anyway, so why not before starting editing source files?

> Hmm, is there any way of getting the paths to global includes platform-independently?

That's the whole point of the build system, isn't it?
That's exactly what `./configure` is supposed to do.

umlaeute:
for what it is worth: your "theoretical" use-case is what is used to build the Debian packages (`soundscaperenderer` (with GUI) and ``soundscaperenderer-nox` (without GUI))

as for compiler-generated files, i have no clear position (though I think they usually do go into the `.gitignore` file)

dvzrv:
@mgeier I'd be totally fine using the VPATH method for my personal use-cases.
The whole ycm topic is a different issue and needs to be persued somewhere else.
Is this otherwise fine for merge or does it require more changes?

mgeier:
I guess `configure.orig` is in there by mistake?

I still think that `compile_commands.json` and `tags` are platform-specific and user-specific things. But I guess as long as they are not too many, they don't hurt too much, so I'll condone it.

dvzrv:
Changed

dvzrv:
I'm not sure why configure.orig ended up in there.
Yeah, compile_commands.json and tags are probably something for another issue.

PR #79, opened 2017-06-05
src/controller.h: Wrapping call to Subscriber::set_source_orientation…

dvzrv:
…() in a call to Scene::get_source_orientation() first, to make the former not segfault, if a non-existent Source is supplied.

### DIFF
--- a/src/controller.h
+++ b/src/controller.h
@@ -1534,10 +1534,12 @@ void
 Controller<Renderer>::orient_source_toward_reference(const id_t id)
 {
   // take reference offset into account?
-  
-  _publish(&Subscriber::set_source_orientation, id
-    , (_scene.get_reference().position - 
-        *_scene.get_source_position(id)).orientation());
+  if (_scene.get_source_orientation(id))
+  {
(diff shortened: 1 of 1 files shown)


mgeier:
I guess the crash happens when dereferencing `*_scene.get_source_position(id)`, right?

What about checking the returned `unique_ptr` for non-NULL before calling `_publish()`?

That will have basically the same effect but it seems a bit less random ...

dvzrv:
@mgeier Yes, you're right.
I just changed and tested this and it works. Eventually the amount of warning messages could be lowered somehow though (now two, in case of trying to position a non-existent source).

In any case, this is ready to be merged.

mgeier:
Thanks.

This is perfectly fine as it is, but what do you think about avoiding a second call to the same function?

Something like this (untested!):

```c
if (auto src_pos = _scene.get_source_position(id))
{
  ...
  (... - *src_pos).orientation()
  ...
}
```

dvzrv:
Good idea! Will try it out in a little bit and report back/push changes.

mgeier:
I've added my suggested change, LGTM!

PR #88, opened 2018-02-12
Improvements for more explicit code

chris-hld:
These commits improve on code that g++ deduced as intended, but other compilers might not (in this case MSVC).

### DIFF
--- a/src/binauralrenderer.h
+++ b/src/binauralrenderer.h
@@ -157,7 +157,7 @@ BinauralRenderer::_load_hrtfs(const std::string& filename, size_t size)
     = std::max_element(transpose.slices.begin()->begin()
       , transpose.slices.begin()->end(), _cmp_abs);
 
-  int index = std::distance(transpose.slices.begin()->begin(), maximum);
+  int index = static_cast<int>(std::distance(transpose.slices.begin()->begin(), maximum));
 
   auto impulse = apf::fixed_vector<sample_type>(index + 1);
   impulse.back() = 1;
@@ -286,7 +286,7 @@ void BinauralRenderer::Source::_process()
   _interp_factor = interp_factor;  // Assign (once!) to BlockParameter
   _weight = this->weighting_factor;  // ... same here
 
-  float angles = _input.parent._angles;
+  auto angles = static_cast<float>(_input.parent._angles);
 
   // calculate relative orientation of sound source
   auto rel_ori = -ref_ori;
--- a/src/configuration.cpp
+++ b/src/configuration.cpp
@@ -154,8 +154,8 @@ ssr::conf_struct ssr::configuration(int& argc, char* argv[])
   conf.path_to_scene_menu = "./scene_menu.conf";
   conf.end_of_message_character = 0; // default: binary zero
 
-  conf.renderer_params.set("decay_exponent", 1);  // 1 / r^1
-  conf.renderer_params.set("amplitude_reference_distance", 3);  // meters
+  conf.renderer_params.set("decay_exponent", 1.0f);  // 1 / r^1
+  conf.renderer_params.set("amplitude_reference_distance", 3.0f);  // meters
 
   conf.auto_rotate_sources = true;
 
--- a/src/rendererbase.h
+++ b/src/rendererbase.h
@@ -120,9 +120,9 @@ class RendererBase : public apf::MimoProcessor<Derived
         , reference_offset_orientation(fifo)
         , master_volume(fifo, 1)
         , processing(fifo, true)
-        , decay_exponent(fifo, params.get("decay_exponent", 1))
+        , decay_exponent(fifo, params.get<sample_type>("decay_exponent", 1))
         , amplitude_reference_distance(fifo
-            , params.get("amplitude_reference_distance", 3))
+            , params.get<sample_type>("amplitude_reference_distance", 3))
       {}
 
       apf::SharedData<Position> reference_position;
--- a/apf/apf/mimoprocessor.h
+++ b/apf/apf/mimoprocessor.h
@@ -484,7 +484,7 @@ APF_MIMOPROCESSOR_BASE::_process_current_list_in_main_thread()
 }
 
 APF_MIMOPROCESSOR_TEMPLATES
-class APF_MIMOPROCESSOR_BASE::Xput : public Item
+class APF_MIMOPROCESSOR_BASE::Xput : public APF_MIMOPROCESSOR_BASE::Item
 {
   public:
     // Parameters for an Input or Output.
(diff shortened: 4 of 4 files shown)


src/rendererbase.h
```diff
@@ -120,9 +120,9 @@ class RendererBase : public apf::MimoProcessor<Derived
         , reference_offset_orientation(fifo)
         , master_volume(fifo, 1)
         , processing(fifo, true)
-        , decay_exponent(fifo, params.get("decay_exponent", 1))
+        , decay_exponent(fifo, params.get("decay_exponent", 1.f))
```
mgeier: Wow, I think this is actually a bug, isn't it?

But your fix also has a bug: `decay_exponent` uses `sample_type`, not `float`. Sure, most of the time it's the same, but the whole point of `sample_type` is that it should also work with `double` (which it practically probably doesn't, but that's another story) or probably even some other type.

The following could work?

```c++
decay_exponent(fifo, params.get<sample_type>("decay_exponent", 1))
```

apf/apf/iterator.h
```diff
@@ -898,7 +898,8 @@ class transform_iterator
     /// Dereference operator.
     /// Dereference the base iterator, use it as argument to the stored function
     /// and return the result.
-    /// @note This is non-const because _f might have state
+    /// @note This could be non-const because _f might have state
+    reference operator*() const { return _f(*_base_iterator); }
```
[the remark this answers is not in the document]
chris-hld: Let's make that a separate PR and discuss this!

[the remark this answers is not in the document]
chris-hld: Ah, yes!
Should I revert this and make a new PR?

src/binauralrenderer.h
```diff
@@ -286,7 +286,7 @@ void BinauralRenderer::Source::_process()
   _interp_factor = interp_factor;  // Assign (once!) to BlockParameter
   _weight = this->weighting_factor;  // ... same here
 
-  float angles = _input.parent._angles;
+  float angles = (float)_input.parent._angles;
```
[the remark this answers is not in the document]
chris-hld: Much better!

src/rendererbase.h
chris-hld: Good point.
It produced just a warning. Doesn't this simply get promoted from int to float in this case?
However, this is later used with `params.get<float>("decay_exponent")`
so it should still be a float type, right?

> decay_exponent(fifo, params.get<sample_type>("decay_exponent", 1.f))

and 
> params.get("amplitude_reference_distance", 3.f))

BTW: controller.h:862 says it throws an error. Maybe that's why?

src/configuration.cpp
```diff
@@ -154,8 +154,8 @@ ssr::conf_struct ssr::configuration(int& argc, char* argv[])
   conf.path_to_scene_menu = "./scene_menu.conf";
   conf.end_of_message_character = 0; // default: binary zero
 
-  conf.renderer_params.set("decay_exponent", 1);  // 1 / r^1
-  conf.renderer_params.set("amplitude_reference_distance", 3);  // meters
+  conf.renderer_params.set("decay_exponent", 1.0f);  // 1 / r^1
+  conf.renderer_params.set("amplitude_reference_distance", 3.0f);  // meters
```
chris-hld: There is a comment in controller.h:862, is this related?

apf/apf/iterator.h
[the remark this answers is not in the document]
mgeier: OK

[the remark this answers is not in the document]
mgeier: As you like.

src/rendererbase.h
mgeier: OK, I had another look and this indeed is a bug!

We didn't notice it before because both `load_scene()` and `_create_spontaneous_scene()` (in `controller.h`) overwrite this value again (and they use the correct invocation of the `get()` function).
So this bug would only have an observable effect if no scene and no sound file is loaded.

> Doesn't this simply get promoted from int to float in this case?

No!
It turns out this is actually quite dangerous ...
The `get()` function is a function template and the target type of the conversion is in this case deduced from the default argument, so it was `int`.
So if the user originally specified `0.5` or `0.0`, this wouldn't be convertible and it would return the `int` value 1 (which would later be converted to a floating-point value).
However, if the user specified `0` on the command line or in the config file (which is a very likely value) everything would have worked fine.

So the best solution is to explicitly specify the template parameter.
After that, it doesn't matter anymore if `1` or `1.0f` or `1.0` is used as default value, because in this case it is indeed automatically converted to the given type.
I would prefer `1`, because there is no need to be more specific.
Or does this produce a warning in your case?
If yes, can you please show it?

src/configuration.cpp
mgeier: Related to what?

chris-hld: Ah no it's fine.

src/rendererbase.h
chris-hld: Ok, now I get it.

chris-hld:
@mgeier This PR should be ready for merge.
Let's discuss the `const` thing separately.

mgeier:
@chris-hld 

> Let's discuss the `const` thing separately.

OK. Can you please create an issue as a reminder (either here or in the APF repo)?

JensAhrens:
@chris-hld Please squash the PR into a smaller meaningful set of commits (or one commit?). We'll then merge it.

chris-hld:
@JensAhrens a single commit should be enough.

src/rendererbase.h
```diff
@@ -120,9 +120,9 @@ class RendererBase : public apf::MimoProcessor<Derived
         , reference_offset_orientation(fifo)
         , master_volume(fifo, 1)
         , processing(fifo, true)
-        , decay_exponent(fifo, params.get("decay_exponent", 1))
+        , decay_exponent(fifo, params.get<sample_type>("decay_exponent", 1))
```
JensAhrens: Should this use 1.0f instead of 1?

src/rendererbase.h
```diff
@@ -120,9 +120,9 @@ class RendererBase : public apf::MimoProcessor<Derived
...
         , amplitude_reference_distance(fifo
-            , params.get("amplitude_reference_distance", 3))
+            , params.get<sample_type>("amplitude_reference_distance", 3))
```
JensAhrens: Same here.

src/rendererbase.h
chris-hld: We had that discussion in src/rendererbase.h above.
It doesn't matter, and @mgeier preferred it this way.
Since I already squashed, this is a lot harder to see now...

src/rendererbase.h
chris-hld: Same as above.

src/rendererbase.h
chris-hld: Long story short:
Here, we explicitly specify the template parameter to `sample_type`.

mgeier: For reference, this is the discussion from above: https://github.com/SoundScapeRenderer/ssr/pull/88#discussion_r168017444

It turns out that `apf::parameter_map` is unsafe and we should fix it some day. I just created #104 as a reminder.

PR #90, opened 2018-02-13
Darker muted sources

chris-hld:
Darkening the color of muted sources.

This makes it so much easier to recognize whether sources are in mute state.


umlaeute:
i haven't tried it, but: isn't the a "lighter colour" (less saturation) the usual way to indicate "disabled" GUI-elements?

mgeier:
@chris-hld A screenshot would probably be helpful.

chris-hld:
screenshot from 2018-02-18 17 07 57
screenshot from 2018-02-18 17 07 35

chris-hld:
Hm, I would say this is pretty clear. What do you think?

chris-hld:
@umlaeute I know what you mean, but this does not touch the saturation. It's not "grayed out".
Just thought it's a nice and simple way. But if you prefer we could do more.

PR #92, opened 2018-03-03
Remove compile-time check for _REENTRANT in cxx_thread_policy.h

### LINKED ISSUE #91: configure.ac needs -D_REENTRANT
On Arch Linux, when building from git source, I get plenty of

```
./../apf/apf/cxx_thread_policy.h:31:2: error: #error You need to compile with _REENTRANT defined since this uses threads!                                                 
 #error You need to compile with _REENTRANT defined since this uses threads! 
```

which lets the bulid fail.
`-D_REENTRANT` needs to be added to CPPFLAGS in configure.ac.

mgeier:
This was removed from the rest of the code base in commit b50218f59ed379ad3645a04bdb37fd72475f5ff4.

Closes #91.

### DIFF
--- a/apf/apf/cxx_thread_policy.h
+++ b/apf/apf/cxx_thread_policy.h
@@ -27,10 +27,6 @@
 #ifndef APF_CXX_THREAD_POLICY_H
 #define APF_CXX_THREAD_POLICY_H
 
-#ifndef _REENTRANT
-#error You need to compile with _REENTRANT defined since this uses threads!
-#endif
-
 #ifndef APF_MIMOPROCESSOR_THREAD_POLICY
 #define APF_MIMOPROCESSOR_THREAD_POLICY apf::cxx_thread_policy
 #endif


chris-hld:
Ah, I was already wondering about this one. I guess it's safe to assume we don't use any old library that needs this?

mgeier:
I guess not.

AFAIK, the option "-pthread" should be used nowadays and that should take care of everything.

mgeier:
But I don't know if that's even true anymore, since C++11 included threading. Probably the compiler should take care of this on its own without needing our help?

dvzrv:
`-lpthread` seems definitely needed (see #95).

chris-hld:
Looks like `-pthread` does everything, and `-lpthread` should not be used.
https://stackoverflow.com/questions/2127797/significance-of-pthread-flag-when-compiling

dvzrv:
valid point. will add it!

PR #94, opened 2018-03-12
add reference offset orientation for BRS Renderer

fietew:
Incorporates the reference_offset for the BRIR rendering.

tested with
```
echo -e '<request><reference_offset><orientation azimuth="0"/></reference_offset></request>\0' | nc localhost 4711
```


JensAhrens:
I think that this is great. @mgeier ?

mgeier:
Yes, this looks fine. I guess it doesn't matter if we use `this->parent` or `_input.parent`.

Issue #96, opened 2018-03-15
travis-ci for gcc and clang


dvzrv:
I'm currently working on a [travis-ci configuration](https://github.com/dvzrv/ssr/tree/travis-integration) (pull request in a bit!).
I'm trying to integrate builds for gcc and clang, but the clang build fails during `./configure`, [trying to make use of asio.hpp](https://github.com/SoundScapeRenderer/ssr/blob/ddc5e5e903e7793cfd0f2172878eeb3fdb54c2bf/configure.ac#L528) (the gcc build doesn't!): [log.txt](https://github.com/SoundScapeRenderer/ssr/files/1817248/log.txt)

I'm a little puzzled as to why this is happening. I can not reproduce this on my machine (clang 5.0.1)!

mgeier:
Is it possible to show the content of `config.log` after the error happens?

dvzrv:
I'll try and cat that to the build!

dvzrv:
Here it is: 
[log-with-config-log.txt](https://github.com/SoundScapeRenderer/ssr/files/1817415/log-with-config-log.txt)

mgeier:
I've come across this annoying `Locale::gettext` thing, but I guess just not installing `help2man` should work, right?

The ASIO error is strange:

```
In file included from /usr/include/asio.hpp:27:
In file included from /usr/include/asio/basic_socket_iostream.hpp:24:
/usr/include/asio/basic_socket_streambuf.hpp:97:21: error: use of undeclared identifier 'steady_timer'
    typename Time = steady_timer::clock_type,
                    ^
```

But `steady_timer` should be defined by ASIO itself ...
There are a lot of preprocessor switches around its definition, probably some of the preprocessor macros are defined differently on `clang`?

chris-hld:
> checking asio.hpp usability... no
checking asio.hpp presence... yes
configure: WARNING: asio.hpp: present but cannot be compiled
configure: WARNING: asio.hpp:     check for missing prerequisite headers?
configure: WARNING: asio.hpp: see the Autoconf documentation
configure: WARNING: asio.hpp:     section "Present But Cannot Be Compiled"
configure: WARNING: asio.hpp: proceeding with the compiler's result
configure: WARNING:     ## ----------------------------------- ##
configure: WARNING:     ## Report this to ssr@spatialaudio.net ##
configure: WARNING:     ## ----------------------------------- ##
checking for asio.hpp... no
[...]

PR #97, opened 2018-03-16
Travis integration

dvzrv:
This adds travis integration. Automated builds for:

- linux (ubuntu trusty based): gcc 4.8.4 (current), gcc6, gcc7 (still fails because of #81), clang 3.8, clang 5.0.0
- macOS (Xcode 8.3, XCode 9.2): gcc and clang (the former being abstracted by llvm) (all have help2man disabled, because on macOS it doesn't have `Locale::gettext` and will fail during build).

Please note: For clang on linux (ubuntu trusty) asio had to be built from source, as the build process is failing with the included version (0.10.1). It is now at 0.12.0.
Also, I'm already building liblo 0.29 (trusty and most other ubuntu versions are still stuck on 0.28) from source in preparation for testing and merging my native OSC interface in the future.

See #96 for reference.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,17 +3,17 @@ language: cpp
 git:
   depth: 3
 
+dist: trusty
+sudo: required
+osx_image: xcode9.2
+
 compiler:
   - clang
   - gcc
 
-matrix:
-  include:
-    - os: linux
-      dist: trusty
-      sudo: required
-    - os: osx
-      osx_image: xcode9.2
+os:
+  - osx
+  - linux
 
 before_install:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -94,7 +94,7 @@ script:
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then make && sudo make install; fi
 
 after_failure:
-  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then ls -alhv /usr/bin/ |grep -E "(clang|gcc|g++)"; fi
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -0,0 +1,106 @@
...
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then ./configure --enable-app-bundle; fi
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then ./configure --prefix='/usr'; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then make dmg; fi
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then make && sudo make install; fi
+
+after_failure:
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then ls -alhv /usr/bin/ |grep -E "(clang|gcc|g\+\+)"; fi
+  - if [[ -f config.log ]]; then cat config.log; fi
+
+branches:
```
mgeier: Is it necessary to limit the branches?

I normally set Travis-CI to only check PRs, so this setting wouldn't matter, right?

dvzrv: @mgeier I think not necessarily, but I wanted to activate it for the feature branches, I'm currently working on. Eventually those can be removed I guess (for now they are useful to me in my clone of the SSR). Having the `master` branch in there is probably never a bad choice though and as long as there is no branch called `travis-integration` or `networking-with-osc`, there shouldn't be any effect in the main SSR repository.

The Travis-CI documentation is sadly not very clear about the "Build pushed branches" feature in the settings. However, from what I understand, if disabled, Travis should not build branches automatically (on commit).

Apart from triggering builds with a pull request ("Build pushed pull requests"), I suggest doing weekly or monthly builds for the master branch though, as the Travis-CI vagrant images are changing constantly (and might break the builds in the process).

Issue #98, opened 2018-03-27
Undefined behavior creating ssr::Connection::Connection()


chris-hld:
Appears with ASIO 1.10.6 and Windows.
When constructing a new Connection in
https://github.com/SoundScapeRenderer/ssr/blob/989568c62dbde7ce060b23576908d758578af2c8/src/network/connection.cpp#L62
there is a segmentation fault thrown deep inside the ASIO library.
I could trace it back to a mutex lock inside the library.
This smells like undefined behavior / race condition.

A simple, yet very effective fix is adding this 
  `std::this_thread::sleep_for(std::chrono::milliseconds(100));`
just before the constructor.

I wanted to discuss this first, because as far as I know there were no problems with the unix library, correct?
Actually I haven't tested it with the latest (github) release. 
But in my opinion this tiny sleep should avoid any potential race conditions.
100 ms are the timeout intervals in the network interface, too.
Any opinions?

mgeier:
Hmmm, this doesn't really sound like a fix ...

Is this a bug in ASIO? Or in the SSR?

I don't remember any issues on Linux or macOS.

I guess it's unrelated, but we should probably use `std::make_shared()` instead of the raw `new`?

chris-hld:
It's in the interplay of both.
The SSR tries to use Connection while it's still in a (ASIO) mutex lock.
This happens during the constructor. I'm also a bit surprised that adding a sleep BEFORE the constructor helps...

`make_shared<Connection>` won't work because the constructor is private.
However, the typedef `pointer` casts it to a std::shared_ptr<Connection>.

chris-hld:
Ok, I think I have a bit more insight about the lock now:
First, I introduced a longer 3 seconds sleep.
Then, consider this python3 test client

```
import socket
TCP_IP = '127.0.0.1'
TCP_PORT = 4711
BUFFER_SIZE = 1024
MESSAGE = b"<request><reference><orientation azimuth='45'/></reference></request> \n"
s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
s.connect((TCP_IP, TCP_PORT))
s.send(MESSAGE)
for t in range(10):
    data = s.recv(BUFFER_SIZE)
    print ("received data:", data)
s.close()
```

The moment I start the python script I get:
`received data: b'<update><volume>0</volume><transport>stop</transport><reference><position x="0" y="0"/><orientation azimuth="45"/></reference></update>\n'
`

[...]

PR #99, opened 2018-05-04
WIP demonstrate behavior creating ssr::Connection::Connection()

chris-hld:
WIP

Try with 
`python3 test_client.py`

See https://github.com/SoundScapeRenderer/ssr/issues/98.

### DIFF
--- a/src/network/connection.cpp
+++ b/src/network/connection.cpp
@@ -29,8 +29,6 @@
 
 #include <functional>
 #include <memory>
-#include <thread>   // std::this_thread::sleep_for
-#include <chrono>   // std::chrono::microseconds
 #include "connection.h"
 #include "publisher.h"
 
@@ -62,7 +60,6 @@ ssr::Connection::pointer
 ssr::Connection::create(asio::io_service &io_service
     , Publisher& controller, char end_of_message_character)
 {
-  std::this_thread::sleep_for(std::chrono::seconds(3));
   return pointer(new Connection(io_service, controller
       , end_of_message_character));
 }
--- a/src/network/connection.cpp
+++ b/src/network/connection.cpp
@@ -29,6 +29,8 @@
 
 #include <functional>
 #include <memory>
+#include <thread>   // std::this_thread::sleep_for
+#include <chrono>   // std::chrono::microseconds
 #include "connection.h"
 #include "publisher.h"
 
@@ -60,6 +62,7 @@ ssr::Connection::pointer
 ssr::Connection::create(asio::io_service &io_service
     , Publisher& controller, char end_of_message_character)
 {
+  std::this_thread::sleep_for(std::chrono::seconds(3));
   return pointer(new Connection(io_service, controller
       , end_of_message_character));
 }
--- a/src/network/server.cpp
+++ b/src/network/server.cpp
@@ -28,7 +28,12 @@
 /// Server class (implementation).
 
 #include <functional>
+#include <thread>
+#include <mutex>
 #include "server.h"
+
+std::mutex accept_mutex;  // protects ssr::Server::start_accept()
+
 
 ssr::Server::Server(Publisher& controller, int port
     , char end_of_message_character)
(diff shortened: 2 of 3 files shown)


mgeier:
I still don't know what's causing the error, but I think I now have a somewhat better understanding of what is going on. I think this is an interesting place:

https://github.com/SoundScapeRenderer/ssr/blob/dd019edf0a6cfec823682a146b8998a46a218f5b/src/network/server.cpp#L59-L68

The time delay you introduced in the `Connection` class happens in the beginning of `start_accept()`.
If you would put the delay between `new_connection->start()` and `start_accept()`, I think it would be easier to understand what happens.

What happens if you remove the call to `start_accept()` here (and remove the time delay)?
This would mean that the network interface would only accept a single connection during its lifetime.
But it should still work with a single client.
I guess this doesn't cause the segmentation fault on Windows, right?

I guess (the preparation for) waiting for the next connection somehow causes the segfault.
I don't know why sleeping a short time seems to fix it ...

But the behavior when sleeping several seconds makes sense to me now: The line

    new_connection->start();

... gets the XML representation of the whole scene and immediately sends it to the new client.
After that, the line

    start_accept();

... sleeps a few seconds before it does anything, and after that it waits for a new connection (which doesn't happen in your current test case).
However, since the whole server runs in a single thread, sleeping while creating a new `Connection` object blocks the whole thread and therefore *all* existing connections. That's while the messages from the first connection only start to appear after the sleep time is over.

Does that make sense?

Can you please show the backtrace leading to the segfault?

mgeier:
AFAIU, `handle_accept()` is called by ASIO when a client tries to connect. To handle this connection, it would be enough to call:

    new_connection->start();

The following call to `start_accept()` is to allow another connection to happen at a later time. And when a second connection happens, `start_accept()` is called again to allow a third connection and so on.

Did you try what happens if you remove the call to `start_accept()`?

I still don't see the race condition. What data is the race about?
I'm not sure but I think every call to `start_accept()` is called from the same thread, isn't it?

chris-hld:
You are absolutely right. It should only be a single server thread. 
Unfortunately, it is very complicated for me to reproduce the error now. I don't have the old Windows machine on which it occurred on hand. Seemed like something that should have been handled by the ASIO library anyway.

Just to keep you in the loop, I upgraded my toolchain to GCC 8.1 and the latest ASIO library. Looks like it's fixed. I'll compile the binaries and make a new release on my side.

Regarding this PR, it's up to you. I will incorporate the mutex lock, because in my perspective it makes sense either way, especially since `handle_accept()` calls `start_accept()` again. And we don't know when and how this might happen.

mgeier:
Thanks for the additional info. I agree that it sounded like a bug in ASIO, let's hope it never comes up again ...

I disagree about the mutex lock, though. Why would you use a lock anywhere in a single-threaded context?

> And we don't know when and how this might happen.

I think we should by now have a quite solid idea about that.
`handle_accept()` should be called by ASIO when a new client tries to connect. At this point, the previous call to `start_accept()` must already have finished, otherwise control wouldn't have been given back to ASIO to call `handle_accept()` in the first place. Therefore, `start_accept()` never runs twice at the same time.

AFAIU, in such a single-threaded asynchronous context, a "callback" function like `handle_accept()` call can be considered atomic. That's also the reason why calling `sleep()` should be avoided, because it blocks the whole asynchronous framework. If there *would be* another thread trying to do something, it would do it during the `sleep()`. But we already saw that's not the case, nothing happens during the `sleep()`.

Also, there is no reason why nothing else should happen between `Connection::create()` and `async_accept()` (which you seem to be trying to avoid with your lock). If anything, each of the calls should be locked separately. But since all this happens in only a single thread, this is not necessary.
If a lock *were* necessary here, it would also be necessary on *every single* call to any ASIO function, which clearly wouldn't make sense, right?

mgeier:
I found some official documentation. I'm not 100% sure, but I think this is called to register the "callback" (or rather "handler"):

https://think-async.com/Asio/asio-1.10.6/doc/asio/reference/io_service/post.html

It says:

>  This function is used to ask the `io_service` to execute the given handler, but without allowing the `io_service` to call the handler from inside this function.
>
> The `io_service` guarantees that the handler will only be called in a thread in which the `run()`, `run_one()`, `poll()` or `poll_one()` member functions is currently being invoked.

chris-hld:
Ah nice, looks like it is designed to be thread safe. OK, then it might have been a bug on the libasio side. Let's hope so.
I agree, the mutex lock is unnecessary in this case. I will remove it in the future.

Issue #85, opened 2017-09-12
native OSC interface


dvzrv:
I conceived a native OSC interface in the course of my thesis.
It uses UDP (as TCP seemed unnecessary overhead) and has a certain notion of a client/server architecture, which allows clients to be subscribed to server instances, while receiving their messages according to certain message levels.
Most of this is still best explained in my thesis itself, as I haven't gotten around writing the manual for the interface (for the official ssr documentation) yet.
However, the interface relies on liblo 0.29 (released last month) and you can try it the following way:

```
git clone https://github.com/dvzrv/ssr
cd ssr
git merge --no-edit origin/networking-with-osc origin/configuration-client-server origin/sclang-workflows origin/tests-osc origin/fix-orient-source-toward-reference origin/reproduction-with-hostnames origin/alien-loudspeaker
cd "$srcdir/${_pkgname}"
./autogen.sh
patch -i "$srcdir/${_pkgname}-qt4.patch"
patch -p1 -i "$srcdir/${_pkgname}-ip-interface-gcc7.patch"
export QT_SELECT=4
export CPPFLAGS="${CPPFLAGS} -D_REENTRANT"
export LIBS="${LIBS} -lpthread"
./configure --prefix=/usr
make

```

The build steps are Arch Linux specific (especially the Qt4 stuff and the things relating to gcc > 7.0), but should be fairly translatable to other Linuxes.
All test cases are done using SuperCollider.

Anyways... as I'm finished, I'll start some form of code cleanup and documentation for the manual and would be happy to see this be merged eventually.

Enjoy

mgeier:
Thanks a lot @dvzrv!

I currently don't have the time to check this out, and the next few weeks I'll be away on holidays ...

I'm always interested in the TCP vs UDP discussion ... do you have some further insights or did you just assume the TCP overhead is bad?
And what's the worst case that can happen if a message is lost or the order of some messages is scrambled?
Or can we assume that this never happens on LANs?

Anyway, I think it would be good to make a new SSR release first and then concentrate on adding new features.

Is there any part of your many branches that should be merged before making a new release?

If yes, it would be great if you could make separate PRs for that!

dvzrv:
@mgeier you're welcome! Hope it will be useful. It was very useful to me already (have been using it for the creation of pieces for quadrophony, etc.).
Enjoy your holidays!

[...]

PR #122, opened 2018-12-15
Networking with osc

dvzrv:
As announced in #85, I now rebased my native OSC interface against `HEAD` and merged all relevant other branches (e.g. `configuration-client-server`, `sclang-workflows` and `osc-tests`).

This requires `liblo > 0.29` and adds quite an extensive framework on how to deal with multiple renderers in a network, acting as clients and servers.
As I haven't added any use-case documentation besides the SuperCollider scripts for testing yet, it's easiest to have a look into my [MT](https://static.sleepmap.de/edu/20170728-MT-David_Runge-A_Networking_Extension_for_the_SoundScape_Renderer.pdf) for further details.

I know this is quite an extensive chunk of work, but I tried hard not to interact with the current code base (e.g. removing other components) and basically stuck to implementing the publisher-subscriber interface for all of this.
I'm quite sure there are better ways of doing certain things in my additions, so please let me know what you think and do review (I know it's a really a lot and I'm sorry about that...).

In any case: I just build this and it works, but I bet there are still some bugs! ;-)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -536,6 +536,13 @@ ENABLE_FORCED([osc-interface], [network OSC interface],
   AC_MSG_RESULT([$have_osc_interface])
 ])
 
+ENABLE_FORCED([osc-interface], [network OSC interface],
+[
+  AC_CHECK_HEADER([lo/lo.h], , [have_osc_interface=no])
+  AC_CHECK_HEADER([lo/lo_cpp.h], , [have_osc_interface=no])
+  AC_MSG_RESULT([$have_osc_interface])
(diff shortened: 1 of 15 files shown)


mgeier:
Thanks a lot @dvzrv!

As discussed in person, we will first try to make the existing network more modular, and then the OSC interface suggested here can be implemented as one of the modules. The other modules will initially be the legacy XML-based protocol and a new, yet to be implemented WebSockets-based protocol. Later, other protocols might be added (or not).

PR #124, opened 2018-12-28
Avoid a few warnings

mgeier:


chris-hld:
Ah, this is good, in c++17 this is actually deprecated and not a warning but an error.
I was messing around with the std::filesystem a bit, which needs c++17 (https://github.com/chris-hld/ssr/commit/95b26d561de575caa26669a10763eaafb1d9de66).
However, it's not ready yet under windows, even the latest gcc8.2 has some serious bugs (doesn't even compile when including filesystem on windows).

mgeier:
>  it's not ready yet under windows, even the latest gcc8.2 has some serious bugs

That's unfortunate. I was planning to use `std::filesystem` from now on ...

Did you use

    LDLIBS += -lstdc++fs

?

chris-hld:
Yes, I did.
The bug happens when including and linking `std::filesystem`, and apparently only under windows/MinGW (only cross compilation affected!):

> usr/lib/gcc/x86_64-w64-mingw32.static.posix/8.2.0/include/c++/bits/fs_path.h:412:55: error: no match for 'operator!=' (operand types are 'std::filesystem::__cxx11::path' and 'std::filesystem::__cxx11::path')
>        else if (__p.has_root_name() && __p.root_name() != root_name())

It is a known bug:
https://sourceforge.net/p/mingw-w64/bugs/737/

PR #128, opened 2019-01-17
TST: Update GCC, clang and Xcode versions on Travis-CI

mgeier:
... in the hope of passing CI in #127.

Does anyone need support for lower versions?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,66 +10,80 @@ env:
 
 matrix:
   include:
+    # linux with gcc 8
+    - os: linux
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
+            - g++-8
+      dist: trusty
+      env:
+        - MATRIX_EVAL="CC=gcc-8 && CXX=g++-8"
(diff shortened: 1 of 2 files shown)


mgeier:
I've removed the tests with GCC on macOS, because `g++` is only an alias for `clang++`. Instead of testing GCC, clang was tested twice.

I tried testing with the actual GCC (version 7), but I got a linker error involving Ecasound. I guess to run the test successfully, we would have to compile Ecasound also with GCC (via Homebrew). This sounded too complicated, so I just removed the GCC tests.

dvzrv:
We could also add ccache to speed things up! [SuperCollider does this](https://github.com/supercollider/supercollider/blob/develop/.travis.yml#L24), successfully.

It seems for OSX this [must be installed specifically](https://github.com/supercollider/supercollider/blob/develop/.travis/before-install-osx.sh#L6) beforehand though.

mgeier:
> We could also add ccache

I don't mind that the builds take a long time, but if somebody makes a PR, we can certainly use `ccache`.

PR #131, opened 2019-01-28
WebSocket-based network interface

mgeier:
This pull request introduces a WebSocket-based network interface for the SSR.

See https://github.com/SoundScapeRenderer/ssr/wiki/SSR-IP-OSC-interface.

It also implements the `ssr-json` sub-protocol, which, not very surprisingly, uses JSON. All messages are JSON arrays, consisting of pairs of items, the first one being the command name and the second one being some corresponding data. One list can contain multiple commands. For examples, see the HTML test client.

WebSocket connections can be made to a specified port (by default `9422`). However, also plain HTTP requests can be made at the same port. For now, this is used to serve an HTML test client that allows to susbcribe to events and monitor the state of the SSR. It also allows to send messages to the SSR for testing the WebSocket interface.

In the future, a proper 3D user interface should be provided instead of the test client, see https://github.com/SoundScapeRenderer/ssr/wiki/Browser-GUI.

Usage:

* Make sure all submodules are available (`git submodule update --init`), compile
* Start SSR (WebSocket server is switched on by default) with a scene of your choice
* Open a web browser and connect to `localhost:9422`

Further changes in this PR:

* some modifications to the "subscribe" functions
* rename the current network interface to "legacy network interface"
* require source IDs to be valid XML IDs (except those auto-created by the renderer)

Questions:

* I've changed string identifiers to use a dash (`-`) instead of an underscore (`_`). This is a backwards-incompatible change. Does anyone think this is a problem?

* I haven't yet implemented "start/stop transport" and "reset tracker" because I couldn't come up with a meaningful JSON syntax for this. Any ideas?

* The sampling rate is now part of the renderer, but the current transport frame is part of the scene. Since there might be multiple renderers in the network but only one scene, there might be different sampling rates, making the transport frame ambiguous. Possible fixes:
  * move sampling rate to the "scene information"
  * change "transport frame" to "transport location in seconds"

  What makes more sense?
  Any other suggestions?

Future Work (not in this PR):

* Allow connecting to another SSR instance as a "follower"
* Allow connecting to another SSR instance as a "leader"
* Reduce the numbers of decimal digits in position coordinates, levels and CPU load to reduce network traffic?


dvzrv:
> * I've changed string identifiers to use a dash (`-`) instead of an underscore (`_`). This is a backwards-incompatible change. Does anyone think this is a problem?

No.

> * I haven't yet implemented "start/stop transport" and "reset tracker" because I couldn't come up with a meaningful JSON syntax for this. Any ideas?

In OSC this was something like `"/transport/state" true` or `"/transport/state" false` and `"/tracker/reset"`. Not sure if this is helpful in a JSON context though.

> * The sampling rate is now part of the renderer, but the current transport frame is part of the scene. Since there might be multiple renderers in the network but only one scene, there might be different sampling rates, making the transport frame ambiguous. Possible fixes:
>   
>   * move sampling rate to the "scene information"
>   * change "transport frame" to "transport location in seconds"
>   
>   What makes more sense?
>   Any other suggestions?

I'm not sure it makes a lot of sense to have renderers in a network with different sample rates (at least that obviously complicates things).
Moving samplerate to the scene information as means to enforce a unified samplerate for all renderers in that network? Sounds not too bad.
Transport frame to transport location in seconds re-calculation for (potentially) changing sample rates (e.g. restart a project with a different sample rate) sounds like we might end up with bizarre rounding errors.

> * require source IDs to be valid XML IDs (except those auto-created by the renderer)

What does that mean?

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "gml"]
 	path = gml
 	url = https://github.com/ilmola/gml.git
+[submodule "rapidjson"]
```
dvzrv: I've already requested a release of gml upstream, so it can be packaged properly.
Rapidjson luckily is already in the repositories (at least on Arch).

We need to keep in mind to build ssr from the system's libs, if they're available.

configure.ac
```diff
@@ -766,7 +773,7 @@ echo "|    Razor AHRS .......................... : $have_razor"
 echo "|    VRPN ................................ : $have_vrpn"
 echo "|"
 echo "| Build with Ecasound support ............ : $have_ecasound"
-echo "| Build with IP interface ................ : $have_ip_interface"
+echo "| Network: legacy/WebSocket .............. : $have_ip_interface/$have_websocket_interface"
```
dvzrv: Shouldn't this be two separate things?

src/Makefile.am
```diff
@@ -26,6 +26,9 @@ AM_CPPFLAGS += -I$(srcdir)/../apf
 ## Add GML path
 AM_CPPFLAGS += -I$(srcdir)/../gml/include
 
+## Add RapidJSON path
+AM_CPPFLAGS += -I$(srcdir)/../rapidjson/include
```
dvzrv: We need to distinguish between system libs (if there are any) and the local ones.

src/api.h
```diff
@@ -154,7 +154,7 @@ struct SceneControlEvents
 
 
 /// Subscribable scene information that cannot be directly changed by clients.
-/// @see Publisher::subscribe_scene_information()
+/// @see SubscribeHelper::scene_information()
```
dvzrv: I really like this btw!

src/Makefile.am
mgeier: The reason why I didn't use the system lib is because Debian stable (https://packages.debian.org/stretch/rapidjson-dev) and Ubuntu xenial (which is the most modern Ubuntu available on Travis-CI, https://packages.ubuntu.com/xenial/rapidjson-dev) have a very old version of RapidJSON.

Do you have an idea how we could have both support for system libraries plus a git submodule as a fallback?

Or should we just require users to download RapidJSON by hand and provide the proper source directory?

The problem is that it is a header-only library, so it will not be installed in a default `include/` directory and I assume it isn't available for `pkg-config`.

configure.ac
mgeier: They are two separate things, just on the same line.

I wanted to keep the number of lines low, so that the summary is fully visible on a 25-line terminal.

But probably that's not important ...?

Would you prefer two separate lines?

mgeier:
@dvzrv Thanks for the review!

I will make the sampling rate part of the scene then.

> > * require source IDs to be valid XML IDs (except those auto-created by the renderer)
>
> What does that mean?

Which part?

I'm currently only allowing IDs that conform to this: https://www.w3.org/TR/xml-names11/#NT-NCName (see https://github.com/SoundScapeRenderer/ssr/pull/131/files#diff-1f84d7860b51690b2964fd6241a307aaR378).

The idea behind this is that IDs like `7` can in some circumstances be used like numeric IDs, because such IDs are otherwise not allowed.
If you want, you can also allow numeric IDs in your OSC interface. For example, you can allow both `/source/7/pos 1.2 0.5 0` and `/source/my-super-source/pos 1.2 0.5 0` (your OSC schema looks different, but I hope you get the idea).

The other part means that auto-generated IDs do intentionally not conform to this limitations. This ensures that a user-specified ID can never accidentally match an auto-generated ID.

src/Makefile.am
dvzrv: Oh, I see (man, those packages are old...).

I'm not quite sure why the Debian package for example doesn't provide pkg-config integration, but at least the headers are in the right place: https://packages.debian.org/stretch/all/rapidjson-dev/filelist

Usually, when installed to the system, rapidjson definitely provides a `pkg-config` integration via `/usr/lib/pkgconfig/RapidJSON.pc` (on Arch at least). Therefore we can use a check, such as `pkg-config --atleast-version 1.0.0 RapidJSON` and default back to the vendored rapidjson, in case the system can't match that requirement.
I assume we have to do some `IFDEF` magic for the includes to make this work, but it should not be super complicated.

configure.ac
dvzrv: Ah, now I get it. Probably was a little tired reading this line.
It makes sense and should be extended for the OSC interface (earliest after LAC... time's to scarce right now).

dvzrv:
Okay. I think I have to read up on that, as I'm not quite sure, what would apply for the OSC interface then.
Does this mean, the renderer created sources are by default a string + `:` and the user created ones without the `:`?

mgeier:
> In OSC this was something like `"/transport/state" true` or `"/transport/state" false` and `"/tracker/reset"`. Not sure if this is helpful in a JSON context though.

It should fit into this scheme (BTW, if you know a better word than `state` please tell me!):

```
["state", {"ref-pos": [-1, 1, 0], "processing": true}]
```

I guess for transport start/stop I could use the same syntax as is used in the other direction (from server to client):

```
["state", {"transport-rolling": true}]
```

But "reset tracker" is trickier, because it has no actual value that has to be conveyed ...

I could try this:

```
["state", {"tracker": "reset"}]
```

... but it's a bit strange, because `"reset"` is the only value that will ever be possible.

I'm also not sure if we will extend the tracker functionality in the future. Probably at some point there could be multiple trackers in one renderer?
Probably we want to choose if the tracker is used for position, rotation or both?
It doesn't have to be a head tracker, probably there are trackers for sources at some point?

mgeier:
> Does this mean, the renderer created sources are by default a string + `:` and the user created ones without the `:`?

For now they are auto-created as `.ssr:` followed by a number:

https://github.com/SoundScapeRenderer/ssr/blob/9bb8eec4c64d81e89656cf7fd76b41d72987a903/src/rendererbase.h#L280

I was thinking about dropping the `:`, to have `.ssr7` instead of `.ssr:7`.

But we can change it to something else in order to be compatible with OSC. Just let me know!

Please note that a user should ideally never see those IDs. Those should be an implementation detail and there is no reason why they shouldn't change between SSR versions.

BTW, I'm planning to have IDs like `.asdf:23` for sources auto-generated with the upcoming new ASDF version. But those can of course also be changed if necessary.

mgeier:
OK, so I've moved the sampling rate from the renderer information to the scene information (2bf121b7ed75d54cdcb67f82fdf24462fdec922f).

With the sampling rate as scene information, it also made more sense to use samples for "transport locate". But since the legacy network interface uses time in seconds, I made two functions to have both options (d1f0d6a06cc7aaff8b156fb586db52272bddd108). I think it's also better to have "frames" and "seconds" explicitly in the function name.

I implemented the missing "transport" messages (e2182967f7ef2c585b6d54eeb8e7e1ff1a9c053a) and a message for resetting the tracker (68a556db4603b2a5cdc3f8e94be78f6adfd5738c).

I also took the liberty of renaming `calibrate_tracker()` to `reset_tracker()` (8bc89dad06e3e193f122d96fe1eb1ad02b92187c), because "calibrating" is a little bit more than we are doing in this function.

Finally, I changed the handling of the third-party libraries GML and RapidJSON (60b3bc24515ee9c5fdea9b89f6452b609e4706d3). Now it is possible to use other versions of these libraries that are installed on the system (and it is not necessary to fetch the Git submodules). If they are in a standard location it works automatically, if not, `CPPFLAGS` can be used to specify a custom include directory. This should make it reasonably simple for package maintainers to use their system packages, while still including a copy of those libraries in the tarball.

PR #132, opened 2019-01-28
Drag and drop

JensAhrens:
This enables drag & drop for audio files and scene files into the GUI. This can be handy for quick testing of something. I also changed the signature of the function `load_scene` in ```controller.h```:

https://github.com/SoundScapeRenderer/ssr/commit/74a356aea93bf354923c2b6e765015ec3dc7674b#diff-1f84d7860b51690b2964fd6241a307aaR131

One can now "add" a scene or audio file to an existing one. This can also be handy if one wants to quickly populate a scene without having to write an asd-file.

Questions are:

1) This change of the signature of `load_scene` back propagates into `publisher.h`: https://github.com/SoundScapeRenderer/ssr/commit/74a356aea93bf354923c2b6e765015ec3dc7674b#diff-7fc5376aafac5f57d0519684ea647e06R61
Does this cause any issues?

2) Adding of a scene is enabled by holding the `Shift` key pressed when dragging and dropping. See the [manual](https://github.com/SoundScapeRenderer/ssr/blob/drag-and-drop/doc/manual/gui.rst#drag--drop-of-scenes). [Any other "adding" action](https://github.com/SoundScapeRenderer/ssr/blob/drag-and-drop/doc/manual/gui.rst#keyboard-actions) is currently done through the `Ctrl`key. Unfortunately, this is not possible in the drag & drop case because of possible (or actual) interference with the file browser. Should we change all "adding" action to use the `Shift`key or accept the inconsistency?

### DIFF
--- a/doc/manual/gui.rst
+++ b/doc/manual/gui.rst
@@ -245,6 +245,12 @@ positioning of sources. The properties ``fixed position``, ``muted`` and
 :ref:`4.3 <source_picture>` to see the complete list of properties
 this dialog shows.
 
+Drag & Drop of Scenes
+~~~~~~~~~~~~~~~~~~~~~
+
+You can also drag & drop scenes (or audio files) into the GUI to open them. Currently, you can only drag & drop one single file.
+
+
 .. _keyboard_actions:
 
 Keyboard Actions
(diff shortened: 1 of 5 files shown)


JensAhrens:
As per offline discussion with Matthias, we postpone adding of files via drag & drop. We only implement opening a new scene for now.

PR #136, opened 2019-01-30
Replace pthreads by std::threads and unify layout

chris-hld:
Switching from pthreads to std::threads. With that the SSR itself is now completely on std::threads.

### DIFF
--- a/src/trackerpolhemus.cpp
+++ b/src/trackerpolhemus.cpp
@@ -48,7 +48,7 @@ ssr::TrackerPolhemus::TrackerPolhemus(api::Publisher& controller
   : Tracker()
   , _controller(controller)
   , _az_corr(0.0f)
-  , _thread_id(0)
+  , _thread_id(std::this_thread::get_id())
   , _stop_thread(false)
 {
   if (ports == "")
--- a/src/trackerpolhemus.cpp
+++ b/src/trackerpolhemus.cpp
@@ -33,7 +33,6 @@
 #include <sstream>   // for std::stringstream
 #include <poll.h>    // for poll(), pollfd, ...
 #include <cassert>   // for assert()
-#include <thread>    // std::this_thread::sleep_for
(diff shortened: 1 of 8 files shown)


src/trackerpolhemus.cpp
```diff
@@ -201,44 +200,42 @@ void
...
 
 void*
-ssr::TrackerPolhemus::thread(void *arg)
+ssr::TrackerPolhemus::_thread(void *arg)
 {
   char c;
   std::string line;
 
-  while (!_stopped)
+  struct pollfd fds;
```
chris-hld: I'm assuming moving this out of the while loop improves performance

chris-hld:
Has anyone access to the sources/headers so that we can at least check that it compiles?

JensAhrens:
Which sources/headers?

chris-hld:
Sorry, just found a link in the docs
https://ssr.readthedocs.io/en/0.5.0/operation.html#head-tracking
I'll try to compile with all head trackers enabled.

chris-hld:
Could include and compile every supported tracker!

mgeier:
Thanks @chris-hld, that's great!

Do you really need the thread ID?

What about using `.joinable()` to check if the thread should be joined?

https://github.com/SoundScapeRenderer/ssr/blob/badf8c00ae24360b93fc2c26d304099fce8b7f78/src/websocket/server.h#L79

Also, the crazy re-direction via `void*` is not necessary anymore, you can directly pass a member function to the `std::thread` constructor, e.g.

https://github.com/SoundScapeRenderer/ssr/blob/badf8c00ae24360b93fc2c26d304099fce8b7f78/src/websocket/server.h#L73

chris-hld:
Thank you for the suggestions, I like these changes a lot!

chris-hld:
I tested the razor tracker.

PR #138, opened 2019-02-18
TST: The brew option --c++11 doesn't seem to be supported anymore on Travis-CI

mgeier:


mgeier:
Does anyone have any concerns about this?
Travis-CI seems to be OK with it.

chohner:
brew removed all options so this is def not needed (anymore)

mgeier:
OK, thanks @chohner!

PR #147, opened 2019-07-17
TST: Install apt and brew packages with Travis addon

mgeier:
There is an error on Travis:

```
checking for SNDFILE... no
configure: error: Package requirements (sndfile >= 1.0) were not met:
No package 'sndfile' found
Consider adjusting the PKG_CONFIG_PATH environment variable if you
installed software in a non-standard prefix.
Alternatively, you may set the environment variables SNDFILE_CFLAGS
and SNDFILE_LIBS to avoid the need to call pkg-config.
See the pkg-config man page for more details.
```

Any ideas why this is happening?

The package `libsndfile1-dev` should be installed with `apt` and the package `libsndfile` should be installed with `brew`.

Am I missing something?


mgeier:
OK, I get it now, it's this problem: https://github.com/travis-ci/travis-ci/issues/3505

mgeier:
Now I'm getting this on Xcode 10.3 and 11:

```
In file included from ssr_binaural.cpp:30:

./controller.h:47:10: fatal error: 'libxml/xmlsave.h' file not found

#include <libxml/xmlsave.h> // temporary hack!

         ^~~~~~~~~~~~~~~~~~
```

On Xcode 10.1 it works fine, though.

shadowfaxster:
I had this when trying to build with make on OSX 10.14. I was using libxml2 installed with brew and the path to the header is not quite right. It worked when I added the right path (libxml2/include/libxml2/) to the $CPPFLAGS environment variable. Maybe it’s something similar.

mgeier:
@shadowfaxster Thanks for the info!

I did some searching and found that this is the problem: https://github.com/Homebrew/brew/issues/5068

I'm currently trying a work-around.

If you know a better solution, please let me know!

Issue #16, opened 2014-05-28
more descriptive error-message for samplerate issues


umlaeute:
i have received a [bug-report for the debian package](https://bugs.debian.org/749529), which i would like to hand over to you:

> SSR fails to start with the error:
>      terminate called after throwing an instance of 'std::logic_error'
>      what(): apf::load_sndfile(): "/usr/share/ssr/default_hrirs.wav" has sample
> rate 44100 instead of 48000!
>      /usr/bin/ssr: line 48: 8444 Aborted $SSR_EXECUTABLE "${OPTIONS[@]}"

i think that ssr could be a bit more descriptive when refusing to use a file.
also, it should be documented somewhere, that ssr does not do any resampling, and that it is the user's task to make sure all files are in the correct format (or alternatively _do_ resampling of the HRIRs)

PR #157, opened 2019-09-23
Generic main function (with exception handling)

mgeier:
See #16.

### DIFF
--- a/src/ssr_aap.cpp
+++ b/src/ssr_aap.cpp
@@ -27,12 +27,10 @@
 /// @file
 /// Main file for the AAP Renderer.
 
-#include "controller.h"
+#include "ssr_main.h"
 #include "aaprenderer.h"
 
 int main(int argc, char* argv[])
 {
-  ssr::Controller<ssr::AapRenderer> controller(argc, argv);
-
-  controller.run();
+  return ssr::main<ssr::AapRenderer>(argc, argv);
 }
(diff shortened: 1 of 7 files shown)


umlaeute (commented):
ha!
i've recently received another [bug-report from a somewhat confused Debian user](http://bugs.debian.org/1029102) that again exposes a segfault instead of printing out a human readable error message.

now i *think* that this commit from 3 years ago fixes the error-handling (though obviously not the error itself), but alas! in Debian we use the last [**release**](https://github.com/SoundScapeRenderer/ssr/releases/tag/0.5.0) of ssr...

umlaeute:
ha!
i've recently received another [bug-report from a somewhat confused Debian user](http://bugs.debian.org/1029102) that again exposes a segfault instead of printing out a human readable error message.

now i *think* that this commit from 3 years ago fixes the error-handling (though obviously not the error itself), but alas! in Debian we use the last [**release**](https://github.com/SoundScapeRenderer/ssr/releases/tag/0.5.0) of ssr...

i probably should open an issue for a new release (bt then: Debian has (kind-of) started the freeze for the next release, so I don't think any release of SSR will make it in time...

mgeier:
Good news: a new release is coming soon:  #331.

The main reason for the delay were complications with the macOS code signing and notarization process. We finally decided to drop the macOS App Bundle from the release, which makes everything much simpler.

PR #164, opened 2019-10-17
add a small type check in command `mod-src` of websocket interface

js00070:
I have been reading the source code of ssr recently. I found a small code defect in websocket interface part. And I just added a simple type check to fix it.


mgeier:
Thanks, good catch!

PR #165, opened 2019-10-21
Add SOFA support to binaural renderer

mgeier:
This uses https://github.com/hoene/libmysofa for loading SOFA files.

Currently, this only works when using a single audio thread (e.g. using the command-line option `--threads=1`).
I hope to find a quick solution to enable multiple threads, see also https://github.com/hoene/libmysofa/issues/80.

SOFA files with delays are not supported. I think it would be quite a lot of work to implement this. But maybe I'm wrong. Anyway, I'm not planning to implement this soon.

The interpolation feature of `libmysofa` (and the "neighborhood" search used for that) is currently not used. I'm not planning to add this in this PR, but we might add it later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ addons:
       - jackd1
       - pkg-config
       - libtool
+      - libmysofa-dev
   homebrew:
     update: true
     packages:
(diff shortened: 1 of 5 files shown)


mgeier:
I've added a check for the number of threads (4b65ac945a04ff04e598f768959ef987867ed3b5), so we can merge this.

I will remove/adapt the check once a good solution for multi-threading has been found.

PR #168, opened 2019-10-24
Travis-CI: use -allowUntrusted

mgeier:
... as suggested by this error message:

    installer: Certificate used to sign package is not trusted. Use -allowUntrusted to override.


mgeier:
Well, this doesn't seem to work at all ...

mgeier:
Another approach: #169

PR #172, opened 2019-10-27
Browser-GUI: Add controls for reference position/orientation

mgeier:
For now, only in the 3D view.


mgeier:
If you want to try it:

1) start the SSR
2) open https://7-216861629-gh.circle-artifacts.com/0/browser-gui/index.html in a browser

PR #184, opened 2019-11-20
CircleCI: Install help2man

mgeier:
... in order to be able to create man-pages for the tarball.


mgeier:
Here's an example tarball which includes the man-pages: https://62-216861629-gh.circle-artifacts.com/0/ssr-0.5.0-102-g39f1be8.tar.gz

PR #216, opened 2020-09-23
Fix typo in ASDF parsing "properties-file" -> "properties_file" 

JensAhrens:


mgeier (approved):

HaHeho:
Nice, glad you found it so easily.

I can confirm that the problem is also fixed for me after rebuilding.

PR #215, opened 2020-09-18
CI: Move from Travis-CI to Github Actions

mgeier:
I also tried using GCC on macOS, but didn't manage to make it work (because of linker errors probably caused by incompatible standard libraries).

If somebody has an idea how to use GCC, please let me know.
If not, it's not a problem, I guess most people use Xcode anyway.

PR #217, opened 2020-09-26
CI: Disable Xcode 10.3

mgeier:
CI failed in https://github.com/SoundScapeRenderer/ssr/actions/runs/273991733.

This might be a temporary glitch, because this has worked before rebase-merging #215.

If the problem doesn't go away on its own in a few days, we should consider merging this PR.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -36,22 +36,21 @@ jobs:
               CC: clang-9
               CXX: clang++-9
           - os: macos-latest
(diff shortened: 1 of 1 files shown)


mgeier:
It looks like Xcode 10.3 isn't really supported anymore on Github Actions.

They will *deprecate* all versions up to (and including) 11.4.0 soon: https://github.com/actions/virtual-environments/issues/1688.

JensAhrens (approved):

PR #253, opened 2021-05-06
Add JACK client name to window title

JensAhrens:
This helps keeping orientation if more than one SSR instance is running.

### DIFF
--- a/data/websocket_resources/ssr-test-client.html
+++ b/data/websocket_resources/ssr-test-client.html
@@ -276,7 +276,8 @@ function resetValues(subscription)
   }
   if (!subscription || subscription === "renderer") {
     [
-      "renderer-name", "processing", "ref-pos-offset", "ref-rot-offset",
+      "renderer-type", "renderer-name", "processing", "ref-pos-offset",
+      "ref-rot-offset",
     ].forEach(id => document.getElementById(id).textContent = "n/a");
 
     document.getElementById("loudspeaker-list").innerHTML = "";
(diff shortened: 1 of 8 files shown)


mgeier:
You should also change this in the HTML test client, see:

https://github.com/SoundScapeRenderer/ssr/blob/f2955bb5859dea76f3c97e40ea4d205cfa071940/data/websocket_resources/ssr-test-client.html#L279

https://github.com/SoundScapeRenderer/ssr/blob/f2955bb5859dea76f3c97e40ea4d205cfa071940/data/websocket_resources/ssr-test-client.html#L420

And it turns out that there is a bug in the FUDI client which you copied (a missing space character). This would probably be a good opportunity to fix this.

JensAhrens:
You mean the blank space in the FUDI subscriber that I removed? 

https://github.com/SoundScapeRenderer/ssr/pull/253/files#diff-eee4360245ccc925cb61f165ce39df06af7fa691b94c69a5f7f613ef20ee59cdL423

I though that it was a mistake that it was there in the first place. So I removed it. Should I put it back?

mgeier:
How do the generated FUDI messages look like?

mgeier:
Since the internal API hasn't yet been in an SSR release, we can still change the names if we want.

I think the "name" was previously meant to be used for both finding out the renderer type and its display in a user interface.

I think two separate strings (as suggested in this PR) are better, but we can probably find better names ...

What about `name` and `type`?

This would also fit with the `--name` command line option, which could then set the "name", while the "type" would be hard-coded.

We can also still change those hard-coded strings if you like.

But we can of course do all this in a separate PR, if you prefer.

PR #255, opened 2021-05-06
CI: build Pd externals

mgeier:
I've once again tried to create the Pd externals with CI, and I think I came quite far this time:

* the Linux externals are built and seem to work on my computer (but maybe not on other Linux systems?)
* the macOS externals are built, but I haven't tested them
* the Windows externals are built, but they don't work yet
  * for now, I've only tried 64-bit, once that works, we might want to try 32-bit as well

I'm using @umlaeute's scripts from https://git.iem.at/pd/iem-ci/-/tree/master/pd-lib-builder, which work great for Linux and macOS, they also seem to work for Windows, but in the end the externals cannot be loaded in Pd (tested on VirtualBox).

This is the error message in the Pd console:

```
ssr_binaural~.dll: The specified module could not be found.
 (126)
```

Any ideas how to make them work?

Can somebody please test the macOS externals?

Downloads: See [below](https://github.com/SoundScapeRenderer/ssr/pull/255#issuecomment-835814231).


umlaeute:
could the non-working-of-windows-externals be related to https://git.iem.at/pd/iem-ci/-/issues/1 ?

mgeier:
Thanks @umlaeute, but that doesn't seem to be exactly it.

But probably this helps getting to the bottom of it:

ssr-dca-dependencies

So `msys-xml2-2.dll` and `pd.dll` seem to be missing.

I also tried to copy them (and their transitive dependencies) manually:

https://github.com/SoundScapeRenderer/ssr/pull/255/files#diff-204c28d32d39f781cb2b1f4df54d3672c5f117bbd02b28ee25fb5918517f69cfR117-R126

But with this, Pd crashed when opening the external.
And it doesn't seem like the right way to do this, does it?

umlaeute:
you should ignore the `pd.dll` (it's provided by Pd at runtime).

running `localdeps.win.sh` multiple times, should eventually copy the `msys-xml2-2.dll` (if it can be found).

mgeier:
Yeah, it looks like libxml2 has a horrific amount of dependencies.

> including things like `msys-libstdc++.dll` and the like which we _probably_ do not want to embed.

I don't think so, since the SSR uses (IIRC) only the C API of libxml2, and nothing C++ related.
There is also a ton of Unicode stuff which we probably don't need.

So what's the solution here?

Should be try to build libxml2 from source on MSYS2, maybe disabling much of the stuff?
I've never built libxml2, it sounds scary.

Or should we switch to a more reasonable XML library, probably [pugixml](https://pugixml.org/)?
Apparently, it has an MSYS2 package: https://packages.msys2.org/package/mingw-w64-x86_64-pugixml

mgeier:
OK, It took me a few hours, but I managed to build libxml2 from source!
It's still not small (about 5 MB), but now all those crazy dependencies are gone.

And the locally built libxml2 is copied by `localdeps.win.sh` without problems.

Now it's time for testing!

mgeier:
I'm now using the custom-built libxml2 for all architectures and I've also compiled libsndfile from source to avoid a bunch of dependencies (because we only need to load impulse responses from WAV files, we don't need compressed formats).

The package got quite a bit smaller, I hope it still works!
Here are the new downloads for testing:

* Deken package: https://github.com/SoundScapeRenderer/ssr/suites/2691058738/artifacts/59309020
* Linux externals: https://github.com/SoundScapeRenderer/ssr/suites/2691058738/artifacts/59309021
* macOS externals: https://github.com/SoundScapeRenderer/ssr/suites/2691058738/artifacts/59309022
* Windows externals: https://github.com/SoundScapeRenderer/ssr/suites/2691058738/artifacts/59309023

mgeier:
I'm merging this, but it would still be nice to get some testing for the generated externals.

PR #257, opened 2021-05-07
Add use cases to manual

JensAhrens:


data/MacOSX/Getting-Started.txt
```diff
@@ -1,9 +1,7 @@
 GETTING (THE SSR) STARTED
 
-You should have a proper version of Jack installed by now (32 or 64 bit - matching the architecture of the SSR build you are using). If not, get it from http://www.jackosx.com/
+You should have a proper version of Jack installed by now (32 or 64 bit - matching the architecture of the SSR build you are using). If not, get it from https://jackaudio.org/ or from Homebrew. You can also obtain QJackCtl from there.
```
mgeier, line 3: It would probably be better to move the content of this whole file to the manual and add a link to it here?

data/scenes/loudspeakers_1-2.asd
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+<asdf>
+  <header>
+    <name>.</name>
+    <description>
+      See the Sec. "Use cases" -> "Multiple Input Multiple Output" for an explanation of what this file does.
```
mgeier, line 6: A link would probably be more helpful.

mgeier (commented):
The impulse responses added in this PR are quite big, wouldn't it be better to move them to somewhere more appropriate, like https://github.com/ssr-scenes or https://github.com/SoundScapeRenderer/listening-test or something like that?

data/scenes/loudspeakers_1-2.asd
JensAhrens: Yes! I moved everything to https://github.com/SoundScapeRenderer/listening-test and put a link there.

data/MacOSX/Getting-Started.txt
JensAhrens: This content is actually in the manual. It might be worth keeping this here so that macOS bundle users are provided with this information directly.

mgeier:
Can you please rebase in order to fix the CircleCI error (see #260)?

data/MacOSX/Getting-Started.txt
mgeier: You are right, I didn't consider that this is included in the DMG.

data/scenes/loudspeakers_1-2.asd
mgeier: Very nice! See my comments at https://github.com/SoundScapeRenderer/listening-test/pull/1.

mgeier:
The WAV files seem to still be part of this PR, is this intentional?

JensAhrens:
- Rebased to master
- Removed the remaining impulse responses

doc/manual/conf.py
```diff
@@ -56,3 +56,18 @@
 latex_documents = [
    ('index', 'SoundScapeRenderer.tex', project, u'SSR Team', 'manual'),
 ]
+
+def gh_link_role(rolename, rawtext, text, lineno, inliner,
+                     options={}, content=()):
+    from docutils import nodes, utils
+    github_url = 'https://github.com/SoundScapeRenderer/listening-test'
```
mgeier, line 63: You are using this role for two different repos.

I guess we'll need two roles now.

doc/manual/index.rst
```diff
@@ -59,4 +59,5 @@ source code and issue tracker
     browser-gui
     network
     ssr-as-library
+    use-cases
```
mgeier, line 62: Since the use cases only use the stand-alone SSR (AFAICT), this section could come before `ssr-as-library`.

Actually, now that I think about it, wouldn't most of this stuff be much easier using the SSR Pd externals?

Admittedly, the BRS renderer wasn't available until a few days ago (#266), but isn't this a game changer?

No multiple SSR instances needed, just a few `[ssr_brs~]` objects. No connecting and disconnecting. No start script needed.

And Windows support! Well, that's quite untested ...

I guess the main difference is that the stand-alone SSR supports head tracking, but apart from that everything else should be better with the Pd externals, right?

AFAICT, the examples don't use the head tracker support of the stand-alone SSR, or do they?

doc/manual/conf.py
JensAhrens: Indeed. I fixed that.

doc/manual/index.rst
JensAhrens: I think that the order is fine. First come all the technical things, then come the `use-cases`. 

The BRS renderer in Pd is indeed cool. I'd say that we should add it (but keep also the current use cases with the stand-alone SSR). I'll take care of that.

No, no head tracking on the multiple-instance SSR. This works only with the Pd patch [tracker_to_4_ssr.pd](https://github.com/SoundScapeRenderer/listening-test/blob/efc11f8642f6174b93c83c0c3610ea4263224671/data_linked_in_ssr_manual/tracker_to_4_ssr.pd) because as soon as an SSR instance reads tracker data, the access to that stream is blocked.

PR #268, opened 2021-06-12
Use std::filesystem, posixpathtools -> pathtools

mgeier:
If GCC7 support is needed, this could probably be added with a few hacks, but I wasn't in the mood to do that.

I've also added some tests, because this stuff is basically impossible to get right without tests.

I haven't tested it with Windows path names. Those theoretically work, but who knows.


mgeier:
CircleCI results: https://app.circleci.com/pipelines/github/mgeier/ssr/478/workflows/804792de-556c-4c8f-876e-aa78f363523e/jobs/481

Doxygen docs: https://481-216861629-gh.circle-artifacts.com/0/doxygen/namespacepathtools.html

PR #281, opened 2022-01-23
Two fixes for the Pd externals

mgeier:
The first problems was that the paths used in the Pd externals had extra quotes around them.
I didn't understand why, until I saw that the `<<` stream operator adds those quotes, see https://en.cppreference.com/w/cpp/filesystem/path/operator_ltltgtgt!

The solution is to simply convert the paths to strings first.
I'm still not sure whether "native" or "generic" paths should be used here, I guess we'll see once this is used on Windows.

The second problem was that the directory of the Pd patch (a.k.a. "canvas dir") was initialized too late.
I thought I'd have to refactor the whole thing to solve this, but actually, it was quite a simple and localized fix.

### DIFF
--- a/flext/ssr_brs~.cpp
+++ b/flext/ssr_brs~.cpp
@@ -51,7 +51,7 @@ class SsrFlext<ssr::BrsRenderer> : public SsrFlextBase<ssr::BrsRenderer>
         }
         std::string file_name = _get(argc, argv);
         apf::parameter_map params;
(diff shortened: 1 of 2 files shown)


AppliedAcousticsChalmersUser (approved):
Works well on macOS 12.1 Monterey

JensAhrens (approved):

PR #292, opened 2022-07-21
Provide CIPIC link from archive.org

mgeier:
The original link doesn't work anymore.


mgeier:
While I'm at it, I've updated a few more URLs.

PR #298, opened 2022-10-29
CI version upgrade

mgeier:
#297 failed, time for a CI upgrade.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -80,7 +80,7 @@ jobs:
     env: ${{ matrix.env }}
     steps:
       - name: Clone Git repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           submodules: true
       - name: Select Xcode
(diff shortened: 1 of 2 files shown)


mgeier:
There seem to be some problems with the Intersense library on macOS, any ideas what's going wrong there?

JensAhrens:
I'm not sure if it is really relevant to support InterSense nowadays. Should we simply remove it?

mgeier:
I don't know if anybody uses the InterSense tracker ... maybe somebody does?

I don't think this is reason enough to completely remove support, because it still seems to work on Linux and older macOS versions. We could just remove it from the newer macOS CI.

I had a look at the `config.log` from the failing CI run, and this is the actual error message:

    ld: malformed universal file: slice content is not mach-o or a static library file '/usr/local/lib/libisense.dylib'

This seems to happen since Xcode 13.3.

mgeier:
It looks like it works fine with Xcode 14 again, so I just removed the Xcode 13.4 check.

PR #302, opened 2022-11-14
Fix CircleCI

mgeier:
There was a strange error when running `./configure`, which was fixed by installing (upgrading?) the `binutils` package.


mgeier:
This is the CircleCI run: https://circleci.com/gh/mgeier/ssr/542

PR #304, opened 2022-11-21
CircleCI: build PDF manual before tarball

mgeier:
I found out that the auto-generated tarball contained a file named correctly like the PDF manual, but it had zero size.

The reason was, quite simply, that the PDF manual has to be built first, and only then can it  be included in the tarball.


mgeier:
The result from the `master` branch should be available here: https://app.circleci.com/pipelines/github/SoundScapeRenderer/ssr

PR #313, opened 2022-12-11
configure: check for pre-built manpages and browser-gui resources

mgeier:
In the past, I have made sure that the tarball contains the manpages and the browser-gui files, but today I found out that this is quite useless, because those files are not used when building the SSR from the tarball!

With the changes in this PR, it hopefully works now.


mgeier:
Here's a freshly built tarball for testing: https://output.circle-artifacts.com/output/job/1f1fff29-c0f4-40f4-b39d-a816acaac7b3/artifacts/0/ssr-0.5.0-249-g4bb7a7c.tar.gz

PR #318, opened 2023-01-01
Manual: use unabbreviated first names in Spors2011 reference

mgeier:
Using a letter followed by a dot seems to lead to strange formatting (the line is turned into an ordered list): https://ssr.readthedocs.io/en/latest/renderers.html#spors2011

Also, all other references have the full first names.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -2,8 +2,12 @@
 
 version: 2
 
+build:
+  os: ubuntu-22.04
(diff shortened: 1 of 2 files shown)


mgeier:
While I'm at it, I've also updated the RTD settings.

Here is the rendered page: https://ssr--318.org.readthedocs.build/en/318/renderers.html#spors2011

PR #325, opened 2023-01-19
Disable `ssr` script

mgeier:
About a decade ago, there was only a `ssr` binary.
Then I created a binary for each renderer type, `ssr-binaural`, `ssr-wfs` etc. and added the `ssr` script for backwards compatibility.

At some point, I wanted to delete this, maybe this time has come?

Two years ago, I have also removed it from the user manual: #236.


JensAhrens (approved):

umlaeute:
why don't you just get rid of the script completely?

put a big fat warning on the download page and into the README, and be done with it (automagically closing #330 along the way)

mgeier:
I thought you like it if users get helpful error messages (see e.g. #16)?

And let's be honest: nobody is reading (much less remembering) warnings in the documentation.

umlaeute:
i do like helpful error messages.
but that doesn't mean that i like error messages for the sake of error messages.

i don't see much help here: the script is basically just saying that it doesn't exist.


as for documentation: sure, i hardly ever read documentation after i first acquainted myself with the software.

however, if things go *horribly* wrong (like: i invoked my favorite program (say... `ssr`) right after I've supposedly installed it, and it this said program cannot be found), then I will start to investigate:
- (as an avid commint user i'm probably typing <kbd>s</kbd><kbd>s</kbd><kbd>r</kbd><kbd>TAB</kbd> anyhow, which will immediately show me the existance of `ssr-binaural` and friends, which *might* make me wonder...)
- with the premise of a  successful build and installation and still not being able to use my software as expected, I *might* be inclined to check the software's documentation.
- and if i **then** find a prominent note telling me that `ssr` has been removed, i'm confident that i will understand and remember this.

mgeier:
> i don't see much help here: the script is basically just saying that it doesn't exist.

Exactly. I think most people who use the SSR for listening tests will use some kind of scripts, and if they use `ssr` in their scripts, those scripts will break. And I think it is much more helpful to see this:

```
$ ./my-script.sh 
The script 'ssr' is not supported anymore, please use the full name:
ssr-binaural, ssr-brs, ssr-vbap, ssr-wfs, ssr-aap, ssr-dca or ssr-generic
```

... instead of this:

```
$ ./my-script.sh 
./my-script.sh: 3: ssr: not found
```

PR #327, opened 2023-01-20
Manual: hide TOC by default on index page, add usage instructions

mgeier:
I thought that the full TOC on the first page might be a bit overwhelming?

And I'm not sure if it helps to give some usage instructions ...

### DIFF
--- a/doc/manual/index.rst
+++ b/doc/manual/index.rst
@@ -48,6 +48,11 @@ source code and issue tracker
 
 ----
 
+.. raw:: html
+
+    <details>
+    <summary>click here to see full table of contents</summary>
+
 .. toctree::
 
     general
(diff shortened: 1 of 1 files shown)


mgeier:
Rendered: https://ssr--327.org.readthedocs.build/en/327/

JensAhrens:
I like the idea, but seeing all layers of sections in the TOC is indeed a lot. It is possible to hide subsections? Never mind if not.

mgeier:
> It is possible to hide subsections?

This is what I originally tried, and yes, it's possible with the `:maxdepth:` option (see https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-toctree).

*However*, this also removed the deeper levels from the sidebar, but I would like to have all levels available there.

Then I tried removing the main TOC completely (with the `:hidden:` option), but I think sometimes it might be useful to see the full TOC, so I didn't really like that.

Another option I explored was using a separate HTML page that only contains the TOC, like Sphinx itself does it: https://www.sphinx-doc.org/en/master/contents.html
I think this isn't too bad, but then I had the idea of initially hiding the TOC on the main page but providing a way to show it on demand. So far, I think this is the best option of everything I tried, but maybe there are more things to try?

Anyway, we can merge this PR now and keep improving on the situation if and when we have further ideas?

PR #330, opened 2023-01-26
Fix exit code of ssr --help/--version, print to stdout

mgeier:
#325 has broken CircleCI, because `make distcheck` seems to check the output of `ssr --help` and `ssr --version`.

### DIFF
--- a/data/ssr
+++ b/data/ssr
@@ -1,10 +1,13 @@
 #!/bin/bash
 
-echo "The script 'ssr' is not supported anymore, please use the full name:"
-echo "ssr-binaural, ssr-brs, ssr-vbap, ssr-wfs, ssr-aap, ssr-dca or ssr-generic"
+case "$1" in
+  # "make installcheck" seems to require a "success" in these cases:
+  --help | --version) FD=1; EXITCODE=0 ;;
+  # ... but we want to signal "failure" in all other cases:
+  *) FD=2; EXITCODE=1 ;;
+esac
 
-# "make installcheck" seems to require a "success" exit code in these cases ...
-case "$1" in --help | --version) exit 0 ;; esac
+>&$FD echo "The script 'ssr' is not supported anymore, please use the full name:"
+>&$FD echo "ssr-binaural, ssr-brs, ssr-vbap, ssr-wfs, ssr-aap, ssr-dca or ssr-generic"
 
-# ... but all other uses should be considered a "failure":
-exit 1
+exit $EXITCODE
--- a/data/ssr
+++ b/data/ssr
@@ -1,4 +1,10 @@
 #!/bin/bash
->&2 echo "The script 'ssr' is not supported anymore, please use the full name:"
->&2 echo "ssr-binaural, ssr-brs, ssr-vbap, ssr-wfs, ssr-aap, ssr-dca or ssr-generic"
+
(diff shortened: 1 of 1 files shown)


mgeier:
CircleCI log: https://app.circleci.com/pipelines/github/mgeier/ssr/627/workflows/cb64beed-4446-455b-bb68-5d6ab34c38e4/jobs/678

umlaeute (changes requested):
shouldn't the warning go to `stderr` instead of `stdout`?

as in:

```
echo "foo" 1>&2
```

mgeier:
Yes they should, but `make distcheck` complains if `--help` or `--version` produce anything on `stderr`, so I used `stdout` for everything.

But then I put in a bit more effort and selected the output depending on the flags: 6cb3d5e9909967f35aedc809e388b86c62a5f53c

umlaeute:
> Yes they should, but `make distcheck` complains if `--help` or `--version` produce anything on stderr, so I used stdout for everything.

then there's someting wrong with `make distcheck`.

umlaeute:
>> but `make distcheck` complains

> then there's someting wrong with `make distcheck.`

do you have any logs that show how `make distcheck` complains?
(it probably is a typo and should read `make check`; but even so i would like to see the complaint/error/..., as i haven't been able to find it yet...)

mgeier:
Yes I do, the build of the current `master` is failing: https://app.circleci.com/pipelines/github/SoundScapeRenderer/ssr/282/workflows/6d3142ed-0089-4e8a-8de0-b3db800b2b88/jobs/288

I might have misinterpreted the logs, though.

I guess that's the relevant part:

```
bad=0; pid=$$; list="ssr"; for p in $list; do \
  case '  ' in \
   *" $p "* | *" ../../../data/$p "*) continue;; \
  esac; \
  f=`echo "$p" | sed 's,^.*/,,;s,x,x,'`; \
  for opt in --help --version; do \
    if "/root/checkout/ssr-0.5.0-272-ge93451a/_inst/bin/$f" $opt >c${pid}_.out \
         2>c${pid}_.err </dev/null \
	 && test -n "`cat c${pid}_.out`" \
	 && test -z "`cat c${pid}_.err`"; then :; \
    else echo "$f does not support $opt" 1>&2; bad=1; fi; \
  done; \
done; rm -f c${pid}_.???; exit $bad
ssr does not support --help
ssr does not support --version
make[3]: *** [Makefile:513: installcheck-dist_binSCRIPTS] Error 1
```

I'm using `make distcheck` here to create the tarball:

https://github.com/SoundScapeRenderer/ssr/blob/e93451abc995a9c9e754a6516002cbfd6c4425f5/.circleci/config.yml#L91

umlaeute:
i see. then there probably is nothing wrong with your `distcheck` target (seeing that all this behaviour comes directly from autotools, presumable when not being run in `foreign` mode).

In this case I would suggest to just completely drop the non-functional `ssr` binary (see also #325)

mgeier:
> presumable when not being run in `foreign` mode

I'm not sure if this is relevant, but we use `foreign` here:

https://github.com/SoundScapeRenderer/ssr/blob/e93451abc995a9c9e754a6516002cbfd6c4425f5/configure.ac#L59-L63

umlaeute:
hmpf. i'm out of ideas for now.

Issue #332, opened 2023-02-01
Test packages, release 0.6.1


mgeier:
We didn't want to widely announce release 0.6.0 yet, because the packaging with `brew` and Deken is quite experimental and should be thoroughly tested first.

Once we have tested the packages, fixed any errors and updated the docs, we should make a new release 0.6.1, which we then should announce to a larger audience.

Here's what needs testing:

## `brew`

With a bit of luck, the SSR can be installed with:

    brew install SoundScapeRenderer/ssr/ssr

I know this is a bit repetitive, but I didn't have a good idea how to avoid this.

As mentioned in https://github.com/SoundScapeRenderer/ssr/pull/331#issuecomment-1412575341, the "bottles" are generated by Github Actions, which currently only supports Big Sur and Monterey, and only Intel.
My hope is that the SSR package can still be installed on other systems, but it will have to be compiled from source, which might take a long time.

It is important to check whether the HTML/JavaScript files for the browser GUI are served correctly.
To test this, start with `ssr-binaural --websocket-server` and then open http://localhost:9422 in a browser (which should show the 3D interface).

## Deken

I have uploaded the package (with a bit of hackery, see https://github.com/pure-data/deken/issues/285), which should cover macOS (Intel), Windows (64bit) and Linux.

I tried it on Linux, found the package, could install it, but then I didn't know how to use it ... just creating `[ssr_binaural~]` didn't work. [UPDATE: I added the `ssr` sub-directory to the search path, now it works]

## Debian package

@umlaeute It would be great if you could have a look if you have time.

I think the main difference to the previous version is that we are now shipping HTML and JavaScript files for the browser GUI.
Those files are included in the tarball, but I guess Debian's policy requires to build them from source? This should be done automatically if `yarn` is installed. It can be requested explicitly with `--enable-browser-gui`.

umlaeute:
> It would be great if you could have a look if you have time.

sure. Debian is currently at the beginning of the freeze cycle for the next release, I might just be able to get it ready in time.

> Those files are included in the tarball, but I guess Debian's policy requires to build them from source?

yes.

> This should be done automatically if `yarn` is installed. 

but we (Debian) do not have network access during build time. That is: we cannot use any 3rd party package manager  (`npm`, `yarn`, `pip`...) to install additional dependencies "from the internet".

umlaeute:
after investigating the situtation, i wonder whether i shouldn't just leave out the browser-GUI from the Debian package for now.

[...]

PR #333, opened 2023-02-03
CI: install libmysofa with brew

mgeier:
Since now we have https://github.com/SoundScapeRenderer/homebrew-ssr, we might as well use it, right?

With a bit of luck, this will also fix the Pd externals for macOS, see https://github.com/SoundScapeRenderer/ssr/issues/332#issuecomment-1416342142.


mgeier:
I tried the external from the CI artifacts (https://github.com/SoundScapeRenderer/ssr/actions/runs/4087475582) on macOS Monterey, and it works great!

PR #335, opened 2023-02-13
Make timeline visible

JensAhrens:
There was an if statement on whether there was enough space for the timeline to fit. It was always untrue so that the timeline stayed hidden. Nowadays, we fix the size of the transport panel in pixels which means that the if statement if not required. I did not delete it but commented it out so that we can bring it back at some point.


mgeier (approved):
On Linux the timeline was also shown before, but with this change it's still shown, so everything is good.

PR #337, opened 2023-02-28
Extend pathtools cross-platform

chris-hld:
Introduce `get_home_dir()` and `normalize_path()`.

### DIFF
--- a/src/configuration.cpp
+++ b/src/configuration.cpp
@@ -192,14 +192,17 @@ ssr::conf_struct ssr::configuration(int& argc, char* argv[])
   load_config_file("/Library/SoundScapeRenderer/ssr.conf",conf);
   // load system-wide config file (Linux et al.)
   load_config_file("/etc/ssr.conf",conf);
-  // load user config file (Mac)
-  fs::path home_dir = pathtools::get_home_dir();
-  fs::path conf_file;
-  conf_file = home_dir / "Library/SoundScapeRenderer/ssr.conf";
-  load_config_file(conf_file.make_preferred().string().c_str(),conf);
-  // load user config file (Linux et al.)
-  conf_file = home_dir / ".ssr/ssr.conf";
-  load_config_file(conf_file.make_preferred().string().c_str(),conf);
+
+  if (auto home_dir = pathtools::get_home_dir(); home_dir != fs::path())
+  {
+    fs::path conf_file;
+    // load user config file (Mac)
+    conf_file = home_dir / "Library" / "SoundScapeRenderer" / "ssr.conf";
+    load_config_file(conf_file.make_preferred().string().c_str(),conf);
+    // load user config file (Linux et al.)
+    conf_file = home_dir / ".ssr" / "ssr.conf";
+    load_config_file(conf_file.make_preferred().string().c_str(),conf);
+  }
 
   const std::string usage_string =
 "Usage: " + std::string(conf.exec_name) + " [OPTIONS] <scene-file>\n";
--- a/src/configuration.cpp
+++ b/src/configuration.cpp
@@ -193,13 +193,13 @@ ssr::conf_struct ssr::configuration(int& argc, char* argv[])
   // load system-wide config file (Linux et al.)
   load_config_file("/etc/ssr.conf",conf);
   // load user config file (Mac)
-  std::string home_dir = pathtools::get_home_dir().string();
-  std::string conf_file;
-  conf_file = pathtools::normalize_path(home_dir + "/Library/SoundScapeRenderer/ssr.conf");
-  load_config_file(conf_file.c_str(),conf);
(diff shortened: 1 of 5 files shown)


src/configuration.cpp
```diff
@@ -193,12 +193,14 @@ ssr::conf_struct ssr::configuration(int& argc, char* argv[])
   // load system-wide config file (Linux et al.)
   load_config_file("/etc/ssr.conf",conf);
   // load user config file (Mac)
-  std::string filename = getenv("HOME");
+  std::string filename = pathtools::get_home_dir().string();
```
mgeier, line 196: Wouldn't it make more sense to keep using a `path` here?
Then the following line could use `/` to append components.

src/configuration.cpp
```diff
@@ -193,12 +193,14 @@ ssr::conf_struct ssr::configuration(int& argc, char* argv[])
...
   filename += "/Library/SoundScapeRenderer/ssr.conf";
+  filename = pathtools::normalize_path(filename);
   load_config_file(filename.c_str(),conf);
```
mgeier, line 199: `filename` could still be a `path` here, which also provides a `.c_str()`  method.

src/pathtools.h
```diff
@@ -77,9 +77,38 @@ inline std::string make_path_relative_to_current_dir(const std::string& path
...
+    std::string p_norm = p_canonical.make_preferred().string();
+    return p_norm;
+}
+
+/** Tries to find a home path, otherwise returns the root path.
+ * @return home path
+ **/
+inline fs::path get_home_dir()
+{
+  fs::path p_home = fs::current_path().root_path();
```
mgeier, line 100: I'm not sure about this fallback case ... why would we return `root_path()`?
This seems wrong to me.

I think we could take inspiration from [root_path()](https://en.cppreference.com/w/cpp/filesystem/path/root_path) and return `path()` as a fallback.

This way, the caller could check if there is a HOME directory or not.

We would of course also have to check this when we are using `get_home_dir()`.

src/configuration.cpp
chris-hld: Unfortunately the cross-compiler doesn't like these implicit path to string conversions...

src/pathtools.h
chris-hld: `root_path` is indeed not the best solution. How about `current_path()`? That's also easy to test and perhaps a better fallback. Say a file can also be located in the installation folder, and could then be found as a fallback

src/configuration.cpp
mgeier: Sorry, I wasn't clear.

I'm not talking about any implicit conversions (those are bad!), I'm talking about never using `std::string` at all.

`get_home_dir()` would return a `path`, which would then be manipulated as such, and in the very end, `.c_str()` would be called on it.

src/pathtools.h
mgeier: > That's also easy to test

I disagree. It seems dangerous, because the current path might change at any point.

> and perhaps a better fallback.

I don't think the current path should be involved, it has nothing to do with the HOME directory.

If no HOME directory is defined, this function should basically return an error. If `path()` is returned, this would mean it didn't work and there is no HOME dir.

chris-hld: Ok, I see! I updated to return `path()` now, in case where it fails

src/configuration.cpp
mgeier: Oh, that's unfortunate. It's interesting how C++ reliably manages to have the least expected behavior by default ... but I guess `string().c_str()` is fine.

src/configuration.cpp
```diff
@@ -193,13 +193,13 @@ ssr::conf_struct ssr::configuration(int& argc, char* argv[])
   // load system-wide config file (Linux et al.)
   load_config_file("/etc/ssr.conf",conf);
   // load user config file (Mac)
-  std::string filename = getenv("HOME");
-  filename += "/Library/SoundScapeRenderer/ssr.conf";
-  load_config_file(filename.c_str(),conf);
+  fs::path home_dir = pathtools::get_home_dir();
```
mgeier, line 196: This could be `path()`. Would that really make sense?

I was expecting the config file only to be loaded if there actually is a HOME directory. If not, nothing should happen, right?

Now it seems like `/Library/SoundScapeRenderer/ssr.conf` is loaded but that has just been loaded before (in line 192), so loading it again doesn't really make sense?

I guess you should check `if (home_dir != fs::path())` or something?

mgeier:
@chris-hld What do you think about my latest comments?

To avoid confusion about what I meant, I have implemented them in ed03a85fb5ef4802a2e8c81545e6c1b42f5bdcdf and e31b1e5a71159dec0c8c9ea2356ffd68cda3733e.

PR #351, opened 2023-06-05
SSR is ready for Windows cross-compiling

chris-hld:
With all the latest changes it seems we are finally super close to having the entire code cross-platform ready, including windows.

I would like to start the discussion about the remaining points:

- [ ] Add guards around missing implementations on WIN32 
- [ ] Build system

Anything missing?

It seems we are only missing some code for head tracker specific implementations, which can be guarded for win32 as shown in this PR. I included the headers that work with the proposed.

The build system seems to be the bigger point here. CMAKE with the MXE environment works very nicely for all platform targets. In theory, MXE should also work with autotools/configure, but I couldn't get it working. Has anyone more experience with this?


mgeier:
Very nice!

> Anything missing?

Did you try building it with CI?

I think it would be good if we could provide binaries, right?

> Ecasound not ported/available under windows. Live inputs work nicely, but it would be great to load audio files. Libsndfile works, and load e.g. HRIRs, so we could perhaps write a fallback audio file reader using libsndfile?

Loading sound files is simple with libsndfile, but the problem is providing the correct data to the audio thread in a realtime-safe manner. That's hard enough, but then we also want to support seeking, which can get very complicated.

The good news is that I have implemented all this for my new ASDF format, see #155.

This will be available very soon. In 3 to 4 weeks, to be exact.

I'm not planning to back-port this to the old ASDF, though.

> The build system seems to be the bigger point here. CMAKE with the MXE environment works very nicely for all platform targets. In theory, MXE should also work with autotools/configure, but I couldn't get it working. Has anyone more experience with this?

I have used MXE in the past, but I recently found that MSYS2 is easier to set up (at least in Github Actions).

I have tried compiling the SSR on MSYS2 some time ago, and I've just rebased and extended this experiment: #352.

mgeier:
I have updated #352, now it builds `.exe` files, but I have the feeling that they won't work just yet. I don't have a Windows system to test them, though.

I think one advantage of MSYS2 over MXE is that I already know how to compile my new dynamic ASDF library (#155), which is implemented in Rust: https://github.com/AudioSceneDescriptionFormat/asdf-rust/blob/7035f6fdf1b52206c80af1bed59a14d6e3fb22d8/.github/workflows/main.yml#L133-L153

It looks like Rust is not yet supported in MXE: https://github.com/mxe/mxe/issues/2499, https://github.com/mxe/mxe/pull/2514.

chris-hld:
Ah that's great!
I am actually already creating an installer https://github.com/chris-hld/ssr-win/releases/tag/v0.6.0-win
I switched to using dynamic linking, which makes more sense since all different SSR executables share a significant portion of dependencies, it makes the installer much smaller (50mb vs 500mb or so).
I am using NSIS for that, the script is in the repo. You can run NSIS on Linux to create the Windows installer.
So far it seems they work flawlessly.

The issue with building on CI is that while I can install pre-build MXE packages (see above), the GCC compiler is stuck at GCC5, which doesn't support the modern CPP features, or at least I don't know how. That's unfortunate, because MXE actually switched to GCC11 as default a while ago, but it seems the Ubuntu packages are just not updated yet.
https://github.com/mxe/mxe/issues/2844

MSYS2 looks easy enough, and seems like a drop in replacement? Have you tried to build with CMAKE? It might set some other options, it took some time to get it running, but so far I didn't have issues with it.

mgeier:
> MSYS2 looks easy enough, and seems like a drop in replacement?

AFAIU, no. MXE is for cross-compilation while MSYS2 runs on Windows.

This would mean that Windows users would be able to install MSYS2 on their local machine and compile the latest SSR code from Git.

Which they cannot do with MXE? Or maybe via WSL?

> Have you tried to build with CMAKE?

No, not yet.

PR #360, opened 2023-06-16
GUI: display reference and reference offset

JensAhrens:
Reference position and rotation and reference offset position and rotation are now properly displayed in the GUI with all renderers.

### DIFF
--- a/src/gui/qopenglplotter.cpp
+++ b/src/gui/qopenglplotter.cpp
@@ -362,32 +362,68 @@ void ssr::QOpenGLPlotter::_draw_reference()
   glPushMatrix();
 
   // translate according to reference position
-  glTranslatef(_scene.get_reference().position.x +
-                   _scene.get_reference_offset().position.x,
-                       _scene.get_reference().position.y +
-                           _scene.get_reference_offset().position.y, 0.0f);
+  glTranslatef(_scene.get_reference().position.x,
+                       _scene.get_reference().position.y, 0.0f);
 
-  glPushMatrix();
-
-  glLoadName(REFERENCEINDEX1); // ID of handle to rotate listener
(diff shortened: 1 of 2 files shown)


mgeier:
This is still wrong. The "offset" is supposed to be relative to the reference position *and orientation*. In other words, the offset should be applied in the local coordinate system of the reference.

In the binaural renderer Qt GUI:

If you set this, everything works fine so far:

    ref pos 1 1; ref rot 45;

But if you then set a reference offset:

    ref offset pos 1 1;

... it looks like it is applied in a global coordinate system.
You can (and should!) compare it with the browser GUI, where the offset is correctly applied in the local coordinate system of the reference.

Once this is corrected, there will still be some awkwardness when rotating the head with the mouse (in the presence of some non-zero offset position), but that's something we can fix at a later time.

In the WFS renderer's Qt GUI, the offset is also wrong in the same sense, plus the reference rotation is not visible anymore.
Also I think it would make more sense to put the rhomb shape to the reference position (as it was before c39b4995271827095087210b950c257c4364ba5f), and not to the offset position.

If you look at the situation before (e.g. at e94a05d2b6056e51d1897fac33598ba6de9215a7), everything work fine with the WFS Qt GUI (except for a rendering bug that I just discovered: #361).

In this case, when moving and rotating the rhomb with the mouse, the behavior is as expected: the loudspeaker system rotates around the rhomb (i.e. the reference position) and the reference offset position moves together with the loudspeakers (assuming that the tracker doesn't move).

JensAhrens:
Check this out now!

mgeier (approved):
Yes, that's it! Now it behaves as I would expect. And the behavior is consistent with the browser GUI as far as I can tell.

PR #364, opened 2023-07-04
DOC: show how to set the Pd path

mgeier:
This was discussed in #332.

Without specifying the path, the externals will not be found, so this is kinda important.


mgeier:
@umlaeute Are those the correct instructions?

Rendered: https://ssr--364.org.readthedocs.build/en/364/use-cases.html#ssr-in-pure-data

umlaeute:
looks good.

i would probably put a `[declare -path ssr]` into the screenshot as well.

it doesn't hurt, and is more prominent than the rather vague hint in the running text.


~~~diff
- for example by creating an object containing [declare -path ssr] (without the brackets) in your patch.
+ for example by creating an object [declare -path ssr] in your patch
~~~

mgeier:
Thanks for the suggestions!

I have updated the screenshot.
I have also added a sentence about sending messages to the external.

JensAhrens (approved):

PR #365, opened 2023-07-04
DOC: mention ASDF v0.4

mgeier:


doc/manual/general.rst
```diff
@@ -213,13 +213,27 @@ Sec. :ref:`Head-Tracking <head_tracking>`.
...
-e.g. movements of a virtual sound source.
-See https://github.com/SoundScapeRenderer/ssr/pull/155 for latest developments.
+Besides pure audio files, SSR can also read the
+*Audio Scene Description Format (ASDF)*
+Currently, two versions of the ASDF are supported.
+
+There is a legacy version of the ASDF [Geier2008b]_,
+which has been supported since the beginning,
+but which can only describe static scenes.
+Recently, support for a new version 0.4
```
JensAhrens, line 223: I think that it should be "Recently, support for a new version 0.4 *was* added ..."

JensAhrens (approved):
It looks good! We should think of a way to showcase the potential of the new asdf version. As of now, there is a short mentioning that dynamic scenes are possible which is quite humble for what that means...

PR #375, opened 2023-08-01
Pd externals: Don't use std::filesystem

### LINKED ISSUE #374: Pd externals are built with `-mmacosx-version-min=10.15` instead of `10.9`
It was brought up in the [Pd mailing list](https://lists.puredata.info/pipermail/pd-list/2023-07/132439.html) that the Externals (installed via Deken) don't work on macOS 10.14 (Mojave).

We intended to use version 10.9 in our CI settings:

https://github.com/SoundScapeRenderer/ssr/blob/1c6c0b3e2b113c928687527abcfad04502f009a7/.github/workflows/pd-externals.yml#L45

However, this seems to be overruled by `-mmacosx-version-min=10.15` for some reason, see for example this log: https://github.com/SoundScapeRenderer/ssr/actions/runs/5489496909/job/14867315885

Any ideas how to fix this?

We probably don't need as old a version as 10.9, but at least 10.14 still seems to be used.

mgeier:
... to support macos < 10.15.

Closes #374.

### DIFF
--- a/.github/workflows/pd-externals.yml
+++ b/.github/workflows/pd-externals.yml
@@ -41,8 +41,6 @@ jobs:
 
   macos:
     runs-on: macos-latest
-    env:
-      cflags: -mmacosx-version-min=10.9
     steps:
       - name: install macOS packages
         run: |
--- a/flext/Makefile
+++ b/flext/Makefile
@@ -14,7 +14,7 @@ cflags += \
        -DNDEBUG \
 
 define forDarwin
-  cflags += -mmacosx-version-min=10.15
+  cflags += -mmacosx-version-min=10.9
 endef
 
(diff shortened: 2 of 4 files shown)


mgeier:
I'm wondering if this will work on Windows, since it uses `/` as a path separator.

mgeier:
It looks like this works on Windows (https://lists.puredata.info/pipermail/pd-list/2023-12/132848.html), but downgrading our external doesn't really help, since the brew-installed dependencies still don't support that older version (https://lists.puredata.info/pipermail/pd-list/2023-12/132847.html).

Users of older macOS versions cannot use the dependencies from `brew` (they might have to compile them from source) and they'll have to compile the SSR externals from source, cherry-picking the commit of this PR (4fd20320f858ffe66a7357b87e426b4e33f9279c).

PR #377, opened 2023-08-22
CI: use fetch-tags for Deken

mgeier:
This doesn't work yet, it needs a new release (> 3.5.3) of the "checkout" action.

See https://github.com/actions/checkout/pull/579

### DIFF
--- a/.github/workflows/pd-externals.yml
+++ b/.github/workflows/pd-externals.yml
@@ -141,7 +141,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           # check out all tags to get proper version in Deken package
-          fetch-depth: 0
+          fetch-tags: true
       - name: Retrieve Linux externals
         uses: actions/download-artifact@v3
         with:


.github/workflows/pd-externals.yml
```diff
@@ -141,7 +141,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           # check out all tags to get proper version in Deken package
-          fetch-depth: 0
+          fetch-tags: true
```
RobertWieczoreck, line 144: @mgeier 
`fetch-tags: true` will not fetch **all** tags, but all tags sitting on the fetched commits.
If you omit `fetch-depth`, the default value `1` will be used, so you only get the tags on that latest commit. For my use-cases it was always enough to specify `fetch-depth: 50` or maybe `100` to be sure the latest tag will be available - as it should be for CI/CD environments imho.

But I already proposed a change to this behavior in https://github.com/actions/checkout/pull/579#issuecomment-1616149745 (see "Suggestion") - maybe I should start another PR as it would also make the meaning of the option more obvious.

mgeier: Thanks @RobertWieczoreck! I understood that https://github.com/actions/checkout/pull/579 will make `fetch-depth` obsolete in my case. Is this not true?

If it's not true, I don't need this PR and I can just keep using `fetch-depth: 0`, which has worked so far.

RobertWieczoreck: I did just describe the behavior (again, it's already described exactly in the PR), so I hoped you would know that ;)
To make it clear again: `fetch-tags` will fetch all tags belonging to the commits that are fetched. With `fetch-depth` omitted, only the latest commit and its tags will be fetched.

My use-case (back then) was, that I only needed the latest tag for automatic Maven/Gradle versioning and didn't want to fetch the whole history using `fetch-depth : 0` (because it was quite large). As we had feature-based CI/CD I could be sure that the latest tag is within the latest 100 commits. My PR allows to use `fetch-depth: 100` together with `fetch-tags: true` to achieve just that: Fetching the latest 100 commits with their tags.

> [..] obsolete in my case.

Depends on the exact nature of "your case". Next time you could provide more information right away.

As I was curious, I briefly looked at the rest of you workflow.
It seems you only need the tags to get a description of the latest reachable tag using `git describe --tags --always` in the "Run deken" step - am I correct?

First, note that this git command will not return the tag name if the tag is not on `HEAD`, but rather a description of `HEAD` relative to the latest tag - see the [docs](https://git-scm.com/docs/git-describe#_examples).
If you want to get the latest tag name only, use `git describe --tags --abbrev=0`.

For the checkout action it depends on the following:
1. Do you know that the latest commit (HEAD) contains the tag?
   - This could be the case if you know that the workflow only runs directly after releasing (so the tag was just created)
2. Do you know that the tag is within the last `X` commits?

If 1. is the case you can just use `fetch-tags: true` without `fetch-depth` as you proposed.
If 2. is true you can use `fetch-depth: X` and `fetch-tags: true` to save some time.

Otherwise I would just keep using `fetch-depth: 0` only, to fetch all history and tags if it doesn't take too long.

You could also omit both options to fetch only the latest commit and execute a `git fetch --tags` to fetch all tags additionally, but I suspect `git describe` will not work then because it needs the history to find the latest tag reachable from HEAD. If you know your version pattern and have separate prefixes or no tags for other branches you still can find the latest tag then using something like this:
```
#!/bin/bash
versionPrefix="v"
latestVersion=$(git tag --list "$versionPrefix*" 2>/dev/null | grep -Po "(?<=$versionPrefix)\d+$" | sort -V | tail -n1 2>/dev/null || echo "not found")
echo "$latestVersion"
```
In this case a simple incrementing version pattern `v1`, `v2`, ..., `v999`, ... is assumed and I stripped the prefix to get only the integer part of the version (to increment it and create the next tag) - but the `grep` part can be easily adapted to other version patterns.

Hope I could help a bit :)

mgeier: > As I was curious, I briefly looked at the rest of you workflow. It seems you only need the tags to get a description of the latest reachable tag using `git describe --tags --always` in the "Run deken" step - am I correct?

Well, not quite. I want to have a unique identifier for the current commit, which is supposed to reduce to the tag name if the current commit has one.

> First, note that this git command will not return the tag name if the tag is not on `HEAD`, but rather a description of `HEAD` relative to the latest tag - see the [docs](https://git-scm.com/docs/git-describe#_examples).

Yes, that sounds exactly like what I want.

> If you want to get the latest tag name only, use `git describe --tags --abbrev=0`.

No, I don't want that, because that would mean that multiple different Deken packages would have the same version name.

> For the checkout action it depends on the following:
> 
>     1. Do you know that the latest commit (HEAD) contains the tag?

No.

>     2. Do you know that the tag is within the last `X` commits?

No.

> If 1. is the case you can just use `fetch-tags: true` without `fetch-depth` as you proposed. If 2. is true you can use `fetch-depth: X` and `fetch-tags: true` to save some time.
> 
> Otherwise I would just keep using `fetch-depth: 0` only, to fetch all history and tags if it doesn't take too long.

Perfect. That's what I'll do.

Issue #380, opened 2023-12-20
CI: Update macOS and Xcode versions ... once again


HaHeho:
https://github.com/HaHeho/ssr/actions/runs/7278435279/job/19832862548 for `macos-11, Xcode_11.7` failed with
```
Error: Your Xcode (11.7 => /Applications/Xcode_11.7.app/Contents/Developer) is too outdated.
Please update to Xcode 13.2.1 (or delete it).
Xcode can be updated from the App Store.
```

Hence, at least that one configuration should be removed. To prevent that we have to update these configurations regularly, maybe we should stick to manually selected macOS versions (currently  `macos-11`, `macos-12`, `macos-13`), but an automatically updated Xcode version, e.g., with https://github.com/marketplace/actions/setup-xcode-version.

Also, Apple Silicon runners may be available now, so a configuration should probably be included?
https://github.com/actions/runner-images/blob/main/images/macos/macos-13-arm64-Readme.md.

HaHeho:
Another issue seems to be with `macos-11, Xcode_13.2.1`, which was interrupted after 6 hours of trying to build many packages from source during the homebrew setup step. Not sure why.
https://github.com/HaHeho/ssr/actions/runs/7278435279/job/19844796502

mgeier:
Thanks for reporting this!

> Hence, at least that one configuration should be removed.

Indeed: #383.

> an automatically updated Xcode version, e.g., with https://github.com/marketplace/actions/setup-xcode-version.

We could try that, but I'm not sure it's worth it.
I don't mind updating the versions from time to time.
But if you think automating it would be better, feel free to make a PR.

> Apple Silicon runners may be available now, so a configuration should probably be included?

I wish, but it seems that it is still limited to paying customers, see https://github.com/actions/runner-images/issues/8439#issuecomment-1755601587.
I didn't try it though ...

> was interrupted after 6 hours of trying to build many packages from source during the homebrew setup step. Not sure why.

Let's hope that was just a temporary glitch.

mgeier:
> Let's hope that was just a temporary glitch.

I don't think it actually was, there's this message in the logs:

```
Warning: You are using macOS 11.
We (and Apple) do not provide support for this old version.
It is expected behaviour that some formulae will fail to build in this old version.
It is expected behaviour that Homebrew will be buggy and slow.
Do not create any issues about this on Homebrew's GitHub repositories.
Do not create any issues even if you think this message is unrelated.
[...]

PR #382, opened 2023-12-21
Remove code deprecations

HaHeho:
I saw a lot of deprecation warnings when building and figured that some of them could be eliminated. I then noticed that all the `sprintf()` deprecations come from the submodule libraries. Hence, only one warning is prevented (`delta`), and one kind of warning that occurs many times is silenced (OpenGL). The latter may be a warning exclusive to macOS (not sure).

I have quickly verified the `QWheelEvent` change (`delta`). The zoom of an `ssr-binaural` scene worked for me as before (tested on macOS). 

To see the other builds succeed, I have temporarily disabled the two failing macOS CI build configurations, as mentioned in https://github.com/SoundScapeRenderer/ssr/issues/380. These should not be merged and treated separately.


mgeier:
Thanks!

Can you please rebase and remove the CI work-around?

PR #381, opened 2023-12-20
Improve Qt5 linking on macOS

HaHeho:
I think this is cleaner to temporarily link Qt5 from homebrew when Qt6 is also installed, with the main reason that the brew link configuration does not need to be corrected and reverted to Qt6 after building.

The GitHub Actions currently fail due to an outdated macOS+Xcode configuration as described in https://github.com/SoundScapeRenderer/ssr/issues/380. I've manually rerun a newer configuration, which was built successfully.

PR #394, opened 2024-03-24
CI: add macos-14

mgeier:
See also #384.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,6 +26,8 @@ jobs:
               CC: clang-9
               CXX: clang++-9
               CXXFLAGS: -fno-builtin
+          - os: macos-14
+            xcode: Xcode_15.4
           - os: macos-13
             xcode: Xcode_15.1
           - os: macos-13
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -99,6 +99,7 @@ jobs:
           )
           brew install ${PACKAGES[@]}
           echo CPPFLAGS="-I$(brew --prefix)/include" >> $GITHUB_ENV
+          echo LDFLAGS="-L$(brew --prefix)/lib" >> $GITHUB_ENV
           # Link keg-only Qt5
           echo "$(brew --prefix)/opt/qt@5/bin" >> $GITHUB_PATH
           echo PKG_CONFIG_PATH="$(brew --prefix)/opt/qt@5/lib/pkgconfig" >> $GITHUB_ENV
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -98,11 +98,9 @@ jobs:
             websocketpp
           )
           brew install ${PACKAGES[@]}
-          # Link keg-only Qt5 according to homebrew info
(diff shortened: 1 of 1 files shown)


HaHeho:
This is very strange what is going on with the `macos-14, Xcode_15.2` instance.

Looking through the logs I've noticed the following based on e.g. the `qt@5` package. `macos-13, Xcode_14.3` places installed homebrew packages into `/usr/local/opt/qt@5`, whereas they are placed in `/opt/homebrew/opt/qt@5/` on `macos-14, Xcode_15.2`.

`/usr/local/` does not seem to exist on the latter, which is the reason for the Intersense lib install as well as the fix to find `qt@5` during `./configure` to fail on `macos-14, Xcode_15.2`.

HaHeho:
`macos-14, Xcode_15.2` ends up being Apple ARM (see https://github.com/SoundScapeRenderer/ssr/actions/runs/8410773165/job/23029617211?pr=394#step:1:11), although Apple Intel is also available for this combination.

For some reason (I guess this is the future), Apple ARM seems to be the default configuration on `macos-14`. If this is the current state, then it will probably stay like this. Apple Intel should be available, but I don't know if `macos-14-large` is available to open-source and for free (right now or ever). See the [list of available runners](https://github.com/actions/runner-images/tree/main?tab=readme-ov-file#available-images).

To fix this, we have to adapt the build pipeline to handle the slightly different environment on Apple ARM, including the different default install path of homebrew packages. However, supporting and testing builds on Apple ARM has been on the wishlist for a while. Free runners are available now, so this seems the right time to do it. I believe to remember that we started to gather information on what changes to the build scripts may be required?!

mgeier:
Without really expecting it to work, I tried `macos-14` for the homebrew bottle: https://github.com/SoundScapeRenderer/homebrew-ssr/pull/30

And it seems to have worked without warnings!

This is without Intersense but it should be with libmysofa.

mgeier:
> The following line on my local systems, Apple x86, and Apple ARM solves the problem `brew link --overwrite qt@5`. I would hope that this could replace the above lines on all macOS builds.

But didn't we change that in the opposite direction in #381?

I tried using `brew --prefix` in 70602451d3e5deb1d4783b06b4604661a6b50eb2, wouldn't this be better?

HaHeho:
> > The following line on my local systems, Apple x86, and Apple ARM solves the problem `brew link --overwrite qt@5`. I would hope that this could replace the above lines on all macOS builds.
> 
> But didn't we change that in the opposite direction in #381?
> 
> I tried using `brew --prefix` in [7060245](https://github.com/SoundScapeRenderer/ssr/commit/70602451d3e5deb1d4783b06b4604661a6b50eb2), wouldn't this be better?

Yes, you're right. I didn't remember, but the outlined advantages to other users, i.e. the force link does not need to be reverted after building, seems reasonable. Although, maybe the `brew` behavior also has changed since then. Seems fine to keep that solution, and the change to `brew --prefix` is good to be Apple Intel vs Apple ARM agnostic.

Btw, in my local build on Apple ARM, I'm stuck at the same point as the CI (`ecasound` not being found after modifying `CPPFLAGS` and `LDFLAGS` like you did).

PR #406, opened 2024-10-03
CI: Update macOS matrix with latest versions

HaHeho:
Mostly updating to the latest available versions and adding `macos-15` which should be available now (configurations based on https://github.com/actions/runner-images/tree/main/images/macos).  Lets see how it goes.

If you want to limit the number of jobs, we could get rid of `macos-13, Xcode_14.3.1` as it is maybe obsolete.


HaHeho:
Looks good now. I don't have a way to test the `macos-15` build as I have not updated yet. But I assume it works.

mgeier:
Thanks!

